PR #5, opened 2016-09-14
fix typos

woodshop:
Fix types that prevent from running.

### DIFF
--- a/wavenet.py
+++ b/wavenet.py
@@ -61,7 +61,7 @@ class WaveNet(object):
         with tf.name_scope('preprocessing'):
             mu = self.channels - 1
             # Perform mu-law companding transformation (ITU-T, 1988).
-            eagnitude = tf.log(1 + mu * tf.abs(audio)) / tf.log(1. + mu)
+            magnitude = tf.log(1 + mu * tf.abs(audio)) / tf.log(1. + mu)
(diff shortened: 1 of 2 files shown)


wavenet.py
```diff
@@ -94,7 +94,7 @@ def _create_network(self, input_batch):
 
             # We skip connections from the outputs of each layer, adding them
             # all up here.
-            total = sum(out)
+            total = sum(outputs[1:])
```
ibab: Why did you leave out `outputs[0]` here?

ibab:
Thanks for fixing these!
I just have a question on one of the lines (see above)

woodshop:
I did not mean to remove `outputs[0]` in the summation. I've fixed the error. I also noticed another type where an unknown variable `i` was being used instead of `step`. I've made these changes and tried to squash into less commits.

wavenet.py
lemonzi: I think this should be `sum(outputs)`, I mis-used a temporary variable.

lemonzi:
These were my fault, sorry about that! I was about to send a patch.

ibab:
This looks good! Thank you!
@lemonzi: No worries. I should set up automated testing.

PR #6, opened 2016-09-14
Fix sampling rate

mecab:
1 << 13 == 8192
1 << 14 == 16384

Sorry if my understand is wrong... (maybe it is because of caring left and right channels so 8192 *2 ?)

### DIFF
--- a/train.py
+++ b/train.py
@@ -73,7 +73,7 @@ def create_vctk_inputs(directory):
         audio_values,
         file_format='wav',
         # Downsample to 16 kHz.
-        samples_per_second=1<<13,
+        samples_per_second=1<<14,
         # Corpus uses mono.
         channel_count=1)
 


lemonzi:
Now that you mention this, I think the standard is 16000, not 16384, isn't it?

ibab:
Setting the sampling rate to 16000 per second is a good idea when trying to reproduce the paper exactly.
@mecab: Could you change the value to 16000?

widdakay:
The dataset is recorded at 48khz, as mentioned here: http://homepages.inf.ed.ac.uk/jyamagis/page3/page58/page58.html

This also is consistent with the naming of "wav48"

mecab:
@ibab 
Sure. I'll make the change for 16000.

@widakkay
Ofcourse 48kHz has better sound, but it will possibly take much longer time to compute, and I doubt it will not improve generated ones' quality comparing the cost paid to process 48kHz.

How about leaving it 16kHz as default, and adding an arguments to allow user to customize the sampling rate?

widdakay:
I somehow missed that this was not setting the sample rate for the file to be decoded at, and rather the sample rate to decimate them too. If it was the opposite, it would result in much slowed down audio. I just checked the documentation, and it does seem to have that effect.

mecab:
I found it has been fixed in 5dbc007 by moving some parameters include this into external files. 🎉 
I'll close the PR.

PR #11, opened 2016-09-14
Read network parameters from an external file

lemonzi:
Loading the model parameters (dilation, etc.) from an external file will make it much easier to reuse the weights from the checkpoint in the audio generation scripts.

There are also other misc edits, mostly related to readability and documentation.

This time I properly checked that everything works :)


layout.json
```diff
@@ -0,0 +1,6 @@
+{
```
ibab: I like the idea of storing these in a file.
I'll make sure to load them from here in the `generate.py` script.

train.py
```diff
@@ -12,40 +20,36 @@
 from wavenet import WaveNet
 
 BATCH_SIZE = 1
-CHANNELS = 256
 DATA_DIRECTORY = './VCTK-Corpus'
-FILTER_WIDTH = 2
 LOGDIR = './logdir'
 NUM_STEPS = 1000
 LEARNING_RATE = 0.10
+NETWORK = './layout.json'
```
ibab: I think `NETWORK_CONFIG` or just `CONFIG` would be a better name, as we're not storing weights in the file.

train.py
```diff
@@ -12,40 +20,36 @@
...
-                        help='Width of the filters to use in the causal '
-                        'dilated convolutions.')
     parser.add_argument('--logdir', type=str, default=LOGDIR,
                         help='Directory in which to store the logging '
                         'information for TensorBoard.')
     parser.add_argument('--num_steps', type=int, default=NUM_STEPS,
                         help='Number of training steps.')
     parser.add_argument('--learning_rate', type=float, default=LEARNING_RATE,
                         help='Learning rate for training.')
+    parser.add_argument('--network', type=str, default=NETWORK,
```
ibab: This could be changed to `--network_config` or `--config`, too.

wavenet.py
```diff
@@ -20,39 +20,40 @@ def _create_dilation_layer(self, input_batch, layer_index, dilation):
...
-        wf = tf.Variable(tf.truncated_normal(
-            [1, self.filter_width, 256, 256],
+        weights_filter = tf.Variable(tf.truncated_normal(
+            [1, self.filter_width, self.channels, self.channels],
             stddev=0.2,
             name="filter"))
-        wg = tf.Variable(tf.truncated_normal(
-            [1, self.filter_width, 256, 256],
+        weights_gate = tf.Variable(tf.truncated_normal(
+            [1, self.filter_width, self.channels, self.channels],
```
lemonzi: Same here.

train.py
```diff
@@ -82,34 +86,38 @@ def create_vctk_inputs(directory):
 
 def main():
     args = get_arguments()
+    logdir = os.path.join(args.logdir, 'train', str(datetime.now()))
```
ibab: TensorBoard seems to handle nested directories well, so I'm 👍 on this.

wavenet.py
```diff
@@ -20,39 +20,40 @@ def _create_dilation_layer(self, input_batch, layer_index, dilation):
         '''Adds a single causal dilated convolution layer.'''
 
         # The filter widths can be configured as a hyperparameter.
-        wf = tf.Variable(tf.truncated_normal(
-            [1, self.filter_width, 256, 256],
+        weights_filter = tf.Variable(tf.truncated_normal(
+            [1, self.filter_width, self.channels, self.channels],
```
lemonzi: That is, assuming that at each layer we map N channels to N channels... Another option would be to give a vector with the same size as `dilations` specifying the number of channels for each layer.

wavenet.py
```diff
@@ -102,7 +106,7 @@ def _create_network(self, input_batch):
 
         return conv2
 
-    def loss(self, input_batch, name='wavenet'):
+    def loss(self, input_batch, add_summary=True, name='wavenet'):
```
ibab: I would expect that `add_summary` would also turn all other summaries on/off.
Let's leave this change out of this PR for now, then we can make all summaries configurable at a later point.

train.py
lemonzi: Yes! I knew it wasn't great but couldn't think of anything better. `NETWORK_CONFIG`? `MODEL_CONFIG`? `WAVENET_CONFIG`? `WAVENET_PARAMS`?

wavenet.py
lemonzi: Oh, I completely forgot about the weight histograms! Yes, let's leave it out.

train.py
lemonzi: TensorBoard is awesome! 😃

layout.json
lemonzi: Ideally it would be a protobuffer, but I think JSON is simpler for this setup.

wavenet.py
ibab: The DeepMind paper doesn't really go into detail about how to treat the channel dimension, so I've simply left it the same throughout the network.
Changing it throughout could definitely make sense, but we should add this at a later point.

train.py
ibab: I like `WAVENET_PARAMS` the most.
The cmd argument could be `--wavenet_params` and the file `wavenet_params.json`.
Then the Python dictionary in the script could be called `wavenet_params` for ultimate consistency.

ibab:
Perfect 👍 
I'm merging this now.

PR #12, opened 2016-09-15
Die gracefully so we don't have to kill the process.

jyegerlehner:


ibab:
Excellent, I was planning on doing this, but never got around to it 👍

Issue #13, opened 2016-09-15
Can't generate samples from checkpoint file


maxhodak:
When I try to run `generate.py` per the readme, I get this:

```
I tensorflow/core/common_runtime/gpu/gpu_device.cc:838] Creating TensorFlow device (/gpu:0) -> (device: 0, name: GeForce GTX TITAN X, pci bus id: 0000:03:00.0)
Restoring model from model.ckpt-250
Traceback (most recent call last):
  File "generate.py", line 86, in <module>
    main()
  File "generate.py", line 66, in main
    feed_dict={samples: window})
  File "/home/ubuntu/jupyter_base/venv/lib/python2.7/site-packages/tensorflow/python/client/session.py", line 710, in run
    run_metadata_ptr)
  File "/home/ubuntu/jupyter_base/venv/lib/python2.7/site-packages/tensorflow/python/client/session.py", line 908, in _run
    feed_dict_string, options, run_metadata)
  File "/home/ubuntu/jupyter_base/venv/lib/python2.7/site-packages/tensorflow/python/client/session.py", line 958, in _do_run
    target_list, options, run_metadata)
  File "/home/ubuntu/jupyter_base/venv/lib/python2.7/site-packages/tensorflow/python/client/session.py", line 978, in _do_call
    raise type(e)(node_def, op, message)
tensorflow.python.framework.errors.InvalidArgumentError: Output dimensions must be positive
         [[Node: wavenet/dilated_stack/layer1/conv_filter/BatchToSpace = BatchToSpace[T=DT_FLOAT, block_size=2, _device="/job:localhost/replica:0/task:0/gpu:0"](wavenet/dilated_stack/layer1/conv_filter, wavenet/dilated_stack/layer1/conv_filter/BatchToSpace/crops)]]
Caused by op u'wavenet/dilated_stack/layer1/conv_filter/BatchToSpace', defined at:
  File "generate.py", line 86, in <module>
    main()
  File "generate.py", line 51, in main
    next_sample = net.predict_proba(samples)
  File "/home/ubuntu/jupyter_base/project/tensorflow-wavenet/wavenet.py", line 154, in predict_proba
    raw_output = self._create_network(encoded)
  File "/home/ubuntu/jupyter_base/project/tensorflow-wavenet/wavenet.py", line 112, in _create_network
    self.dilation_channels)
  File "/home/ubuntu/jupyter_base/project/tensorflow-wavenet/wavenet.py", line 51, in _create_dilation_layer
[...]

PR #15, opened 2016-09-16
Direct wav file output from command line

mecab:
This PR enable the users to have the result directly from command line.
Includes README updates (welcome to fix the writing)

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,6 +53,8 @@ TensorFlow 0.10 and the current `master` version are supported.
 In addition, the `ffmpeg` binary needs to be available on the command line.
 It is needed by the TensorFlow ffmpeg contrib package that is used to decode the audio files.
 
+You will need `scipy` if you want to have the result as a wavfile directly from command line.
+
 ## Training the network
 
 The [VCTK corpus](http://homepages.inf.ed.ac.uk/jyamagis/page3/page58/page58.html) is currently used.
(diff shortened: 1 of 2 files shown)


adroit91:
Are you even able to generate audio from generated models? There's an issue being discussed about that not working! [https://github.com/ibab/tensorflow-wavenet/issues/13](url)

ibab:
Yeah, we need to fix #13 first, then I can merge this.
Theoretically, we could also use the TensorFlow ffmpeg module to save the files, but I'm fine with using `scipy`.

mecab:
I managed to have a "valid" wav file by letting `dilations` be `1` as had mentioned by @ibab (https://github.com/ibab/tensorflow-wavenet/issues/13#issuecomment-247456622)
and avoiding to use GPU in generation to avoid `CUDA_ERROR_ILLEGAL_ADDRESS`.  But it was almost white-noise. I'm not sure whether the result was wrong or the iteration was not enough.

I agree to merge this after fixing #13.

ibab:
@mecab: I've just found the origin of the `CUDA_ERROR_ILLEGAL_ADDRESS` bug, so generating audio files should work in a moment.
What was the loss that you managed to reach before generating?

mecab:
Yes. We can use TensorFlow's FFmpeg. But it makes the dependency outside python (i.e., we need to install FFmpeg). In some shared computational environment, installing FFmpeg to the machine without `sudo` could be a pain.

So I'm for to use `scipy` (or other closed-in-python module) to handling audio, but it's up to you.

mecab:
@ibab 
Super! I'm looking forward to seeing it.
The loss was almost 3, but I used custom dataset so I'm not sure it can be a good reference...

ibab:
Duh, I thought that the bug was a missing colon in the line that crops the previously generated waveform to a smaller window, but it seems to be something else and both errors just happened to be triggered at 256 elements.

ibab:
As audio generating is working properly now, I'll go ahead and merge this.
Thanks a lot for adding this feature!

PR #16, opened 2016-09-16
Rich output

mecab:
Shows #step and processing time on the output in the training.
It will help estimating ETA.


ibab:
This is nice to have. Thank you!

PR #20, opened 2016-09-16
Cast output for workaround of CUDA_ERROR_ILLEGAL_ADDRESS in generation

mecab:
Workaround of #13


mecab:
Fixes #13

ibab:
Cool, could you add a comment explaining that this is a workaround?

mecab:
Right. It is actually not a fix...

Fixed the comment. Could you review?

ibab:
Thanks 🎉

PR #23, opened 2016-09-17
Fix spelling error

ZECTBynmo:
Sorry to submit a PR for such a small issue, but I found what appears to be a spelling error while looking through comments. Thanks for the excellent work


ibab:
Thank you! Every pull request counts :)

PR #27, opened 2016-09-17
Ensure the model is saved at the last iteration.

mecab:


ibab:
Yeah, this is a good fix 👍

PR #28, opened 2016-09-17
Remove colons from the datestring to have logdir

mecab:
It is to avoid the path issue as follows on restoring checkpoint later.

```
tensorflow.python.framework.errors.UnimplementedError: File system scheme `{{ incomplete_path }}` not implemented
```

where `{{ incomplete_path }}` is like: `logdir/train/2016-09-18T01` (terminated before the colon).

### DIFF
--- a/train.py
+++ b/train.py
@@ -86,7 +86,7 @@ def create_vctk_inputs(directory, sample_rate=16000):
 
 def main():
     args = get_arguments()
-    datestring = str(datetime.now()).replace(' ', 'T')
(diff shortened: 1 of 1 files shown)


ibab:
Yeah, that's a good idea.
This might also warrant a bug report to TensorFlow.

mecab:
The path issue is confirmed on 

- Mac OS X 10.11.6
- Python 3.5.2
- Tensorflow 0.10.0

I also changed removed millisec. part in the string for simplify. Better format can be proposed.

PR #34, opened 2016-09-18
Add alternative VCTK corpus hosted site link

chanil1218:
Provided VCTK corpus hosted site may be too slow to complete download in some countries. 
Add another hosted site for concerning server status of hosted site and improving stable downloadability on required dataset.


ibab:
This is useful, thanks!

PR #37, opened 2016-09-18
Unit test for time shifting.

jyegerlehner:
I felt uneasy about the padding razzle-dazzle that the causal conv does, and whether it would end up shifting the output of the filter relative to the input. I know ibab's first test sort of tests this, but only if you believe his trick of hacking the last 4 elements off the result of the numpy convolve output. Turns out he was right, of course. But I wanted to see it for myself :)

If you think this test is just noise I'll close. I think it's conceivable that a future change could break this behaviour so I would feel better to leave it in.

### DIFF
--- a/test/test_causal_conv.py
+++ b/test/test_causal_conv.py
@@ -22,3 +22,35 @@ class TestCausalConv(tf.test.TestCase):
 
         self.assertAllEqual(result, ref)
 
+    # Test that the causal convolution does not shift the output in time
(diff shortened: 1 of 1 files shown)


ibab:
Yeah, the causal conv code is a bit complicated and prone to errors, so it's nice to have more tests.
Can you rebase this on `master` so travis can have a go at running the tests?

ibab:
Awesome, travis is happy :)
Thanks!

PR #38, opened 2016-09-18
threaded audio loader with librosa

genekogan:
This swaps ffmpeg with librosa via a threaded class that continuously adds audio files to the queue. It is mostly based on [this examples](https://indico.io/blog/tensorflow-data-input-part2-extensions/). 

Also of note: there is a function `iterate_through_directory` which will recursively scan any directory for audio files and return them all with label `0`. then there is also a function `iterate_through_vctk' which is the same thing except specific to the VCTK dataset and returns the associated speaker labels with each file. This is the default function used in the script.

It trains for me and produces the expected output, but another set of eyes to verify this would be very helpful.

still to-do: use librosa instead of scipy to write the audio to disk (though this isn't critical)


train.py
```diff
@@ -124,51 +141,51 @@ def main():
...
     finally:
-      coord.request_stop()
-      coord.join(threads)
-
+        pass
+        
 
 if __name__ == '__main__':
-    main()
+    main()
```
ibab: There's an extra space at the end here.

train.py
```diff
@@ -124,51 +141,51 @@ def main():
...
     init = tf.initialize_all_variables()
     sess.run(init)
 
-    # Load and enqueue examples in the background.
-    coord = tf.train.Coordinator()
-    qr = tf.train.QueueRunner(queue, [enqueue])
-    qr.create_threads(sess, coord=coord, start=True)
-    threads = tf.train.start_queue_runners(sess=sess, coord=coord)
+    # start the tensorflow QueueRunner's
+    tf.train.start_queue_runners(sess=sess)
```
ibab: I think we might still want to use a coordinator here to shut down all threads in a clean way.
We can even pass the coordinator to our custom thread runner and query the `coord.should_stop()` method to find out whether we should stop it.

train.py
```diff
@@ -124,51 +141,51 @@ def main():
...
-              writer.add_summary(summary, step)
-
-          duration = time.time() - start_time
-          print('step %d - loss = %.3f, (%.3f sec/step)' % (step, loss_value, duration))
-
-          if step % 50 == 0 or step == args.num_steps - 1:
-              checkpoint_path = os.path.join(logdir, 'model.ckpt')
-              print('Storing checkpoint to {}'.format(checkpoint_path))
-              saver.save(sess, checkpoint_path, global_step=step)
+        for step in range(args.num_steps):
```
ibab: Thanks for cleaning up the indentation here!

ibab:
Btw, if you want travis to pass, you can rebase to the newest `master`, where I've managed to get it to work.

genekogan:
cool!

not entirely sure if i'm doing this correctly. does this look right, reinserting the queuerunner? the CustomRunner still needs to be explicitly started?

```
    # Set up session
    sess = tf.Session(config=tf.ConfigProto(log_device_placement=False))
    sess.run(init)

    # Load and enqueue examples in the background.
    coord = tf.train.Coordinator()
    qr = tf.train.QueueRunner(custom_runner.queue, [custom_runner.enqueue])
    qr.create_threads(sess, coord=coord, start=True)
    threads = tf.train.start_queue_runners(sess=sess, coord=coord)

    # start our custom queue runner's threads
    custom_runner.start_threads(sess)

    # Saver for storing checkpoints of the model.
    saver = tf.train.Saver()
```

ibab:
I think the `QueueRunner` is not necessary in this case.
It just runs an op repeatedly, until the Coordinator tells it to stop, but in our case we want to be able to do more than just running a tensorflow op.

We can pass the Coordinator into the CustomRunner in some way and then break inside the `thread_main` function if `coord.should_stop()` returns `True`.
That way the script won't keep running if there's an exception.

genekogan:
have a look, i think i've made the requested changes.

ibab:
Looks great, thanks!

PR #39, opened 2016-09-19
Let there be biases

jyegerlehner:
Seems to be working a bit better with these.

screenshot from 2016-09-18 20 41 17

The orange line is without biases, and blue line is same model with the biases.

### DIFF
--- a/wavenet.py
+++ b/wavenet.py
@@ -44,12 +44,20 @@ class WaveNet(object):
             [self.filter_width, in_channels, dilation_channels],
             stddev=0.2,
             name="filter"))
+        biases_filter = tf.Variable(tf.constant(0.0, shape=[dilation_channels]),
+                                    name="filter_biases")
+
         weights_gate = tf.Variable(tf.truncated_normal(
             [self.filter_width, in_channels, dilation_channels],
             stddev=0.2, name="gate"))
+        biases_gate = tf.Variable(tf.constant(0.0, shape=[dilation_channels]),
+                                  name="filter_biases")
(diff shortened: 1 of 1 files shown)


ibab:
This is interesting. There don't seem to be biases in the original paper.
Could you turn these into an optional parameter in the `wavenet_params.json` file?
As in `use_biases: true`.

jyegerlehner:
> There don't seem to be biases in the original paper.

Well, to me it appears to be a detail they didn't go into. One of those "it goes without saying" things. Who ever builds a neural net without biases? It would be tedious to put that verbiage in the paper IMO.

> Could you turn these into an optional parameter in the wavenet_params.json file?

OK

ibab:
> Well, to me it appears to be a detail they didn't go into. One of those "it goes without saying" things. Who ever builds a neural net without biases? It would be tedious to put that verbiage in the paper IMO.

That's actually a good point.
Thinking about this, at least the 1x1 layers should definitely have biases, and they should be beneficial for the causal convolutions as well.
I'm pretty convinced that I've interpreted the paper too literally in this regard.

jyegerlehner:
I've re-run code where the biases are disabled and I get the same results as the current master branch. I think my loss plot above is wrong: the difference does not apear to be due to biases. I would guess due to some other change I pulled in to rebase on master. I'm closing this until I figure out what's going on.

jyegerlehner:
Superceded by https://github.com/ibab/tensorflow-wavenet/pull/57

Issue #24, opened 2016-09-17
Add unit tests


ibab:
I'd like to add unit tests before extending the implementation further.
That should prevent us from pushing non-working versions of the code in the future.
TensorFlow has a nice API for tests, which I'd like to use for this: https://www.tensorflow.org/versions/r0.10/api_docs/python/test.html#testing

multivac61:
I have started implementing unit tests - https://github.com/multivac61/tensorflow-wavenet/blob/unittests/test.py

@lemonzi has suggested that code that is independent on hyperparameters (_preprocessing etc) be moved outside the WaveNet class. That makes testing on them easier and it leads to better code structure.

lemonzi:
Yep! I think we could move out `_preprocess` (which is actually `encode`), `decode`, `dilated_conv`, and maybe something more.

ibab:
Yeah, we can create a `wavenet_ops.py` or just `ops.py` file that we can move all independent functions to.

ibab:
I've added a first unit test at [`test/test_causal_conv.py`](https://github.com/ibab/tensorflow-wavenet/blob/0a60d90e5162de222601214e2e9998ba4b6f515f/test/test_causal_conv.py).

lemonzi:
It seems we have decent unit tests for now. Pieces not covered yet should be listed in Projects or in the Wiki.

PR #43, opened 2016-09-19
Add net training test.

jyegerlehner:
Pertaining to [issue 24](https://github.com/ibab/tensorflow-wavenet/issues/24), this adds a test that trains the network to learn a simple audio clip (3 tones forming a chord, at least that's what it is supposed to be). The idea is to provide broad coverage of the code and sanity check that the WaveNet is trainable. 

It runs in a couple seconds of seconds on my machine, but might not be Travis-friendly since I think it has to run on the CPU there. 

If you uncomment the print statement you can see the loss drops very close to zero, so it has successfully overfit to the 0.10 second long clip.

It would be interesting to save the generated audio (and training audio) out to .wav file to see how well it sounds.

Edit: build fails at moment because travis can't find librosa.

### DIFF
--- a/ci/install.sh
+++ b/ci/install.sh
@@ -9,6 +9,7 @@ conda update -q conda
 conda create -q -n test python=$TRAVIS_PYTHON_VERSION numpy scipy
 source activate test
 pip install nose
+pip install librosa
 
 pip install https://storage.googleapis.com/tensorflow/linux/cpu/tensorflow-$TENSORFLOW-cp35-cp35m-linux_x86_64.whl
 
--- a/test/test_net.py
+++ b/test/test_net.py
@@ -2,7 +2,7 @@
 import tensorflow as tf
 import numpy as np
 from wavenet_ops import time_to_batch, batch_to_time, causal_conv
-from train import WaveNet, get_arguments
+from wavenet import WaveNet
 import json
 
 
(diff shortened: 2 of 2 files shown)


ibab:
@jyegerlehner: There was a bug in the ci install script. If you rebase on master, the tests should now run correctly.

ibab:
Can you replace the use of `get_arguments` with default values? (Or maybe the idea was to use the current default args?)
The problem is that this will give an error when using `nosetests` to run the tests, as the `nosetests` parameters become the ones parsed by `get_arguments`.

jyegerlehner:
Now get_arguments() is failing. I see this failure locally if I run 
```
sh ci/test.sh
```
but succeeds if I run 
```
python ci/test_net.py
```
I'm somewhat new to python and know nothing about argparse, so will have to flounder around a bit to figure this out. Or maybe I'll just rip out the get_arguments() call and put some hard-coded params in.

ibab:
> Or maybe I'll just rip out the get_arguments() call and put some hard-coded params in.
Sorry, that's what I meant with default values :)
argparse interferes with the testing framework, and putting hard-coded values in the test should be fine.

jyegerlehner:
> Sorry, that's what I meant with default values :)

Ah my fault; I posted before I saw that comment (needed to refresh my browser). It was perfectly clear. Sure, that's the easiest way to go. At the expense of a bit of lost coverage: I was thinking if we use the same defaults as the rest of the software, then this test does some sanity checking on them.

ibab:
This is a really neat test, thanks for implementing it!

larsr:
Running the test 1200 rounds or more instead of 50 rounds does not converge. The error jumps up to 4 or five and stays there...  I've noticed similar things when training on the voice data, so perhaps there is some problem.  Nice to have it manifest itself so quickly in a test, though.

```
@@ -57,9 +57,9 @@ class TestNet(tf.test.TestCase):
         with self.test_session() as sess:
             sess.run(init)
             initial_loss = sess.run(loss)
-            for i in range(50):
+            for i in range(5000):
                 loss_val, _ = sess.run([loss, optim])
-                #print "i: %d loss: %f" % (i, loss_val)
+                if i%10==0: print("i: %d loss: %f" % (i, loss_val))
 
         # Sanity check the initial loss was larger.
         self.assertGreater(initial_loss, max_allowed_loss)

```

ibab:
That's a nice find.
I haven't encountered this myself when training yet, but I do see it when increasing the number of steps in this test (both on the GPU and CPU).

PR #44, opened 2016-09-19
add librosa to readme

genekogan:


ibab:
Thanks! This pull request is also showing your previous commits. I think this is because you've committed the last change on top of your previous changes, instead of master.
But it should be fine if I squash the changes.

genekogan:
yeah sorry about that, will sync with master before the next one.

PR #46, opened 2016-09-19
Implement checkpoint loading

### LINKED ISSUE #35: Continuity on trained model
It would be great continue learning from the someone's trained model or paused mine.

mecab:
Will close #35. The latest checkpoint in `--logdir` can be restored.

This PR will change the behavior of `--logdir`.
 - With `--logdir`, `--logdir` is directly used as `logdir`.
 - Without `--logdir`, use `LOG_DIR/train/{datestring}` is used as `logdir` as before.

Possible concern is that someone may want to use `SOMEWHERE_THEY_NEED/train/{datestring}` for their first training (i.e., they need the app to make dated directory, but not under default `LOG_DIR`).

The solution is add a flag like `--default_logdir_root` to change `LOG_DIR` part, but it could the flags be complicated.

### DIFF
--- a/train.py
+++ b/train.py
@@ -5,6 +5,8 @@ which can be freely downloaded at the following site (~10 GB):
 http://homepages.inf.ed.ac.uk/jyamagis/page3/page58/page58.html
 '''
 
+from __future__ import print_function
+
 import argparse
 from datetime import datetime
 import glob
@@ -30,7 +32,7 @@ LOGDIR = './logdir'
 NUM_STEPS = 4000
 LEARNING_RATE = 0.02
 WAVENET_PARAMS = './wavenet_params.json'
-
+STARTED_DATESTRING = "{0:%Y-%m-%dT%H-%M-%S}".format(datetime.now())
 
 def get_arguments():
     parser = argparse.ArgumentParser(description='WaveNet example network')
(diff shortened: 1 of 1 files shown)


train.py
```diff
@@ -42,7 +44,10 @@ def get_arguments():
                         help='Whether to store advanced debugging information '
                         '(execution time, memory consumption) for use with '
                         'TensorBoard.')
-    parser.add_argument('--logdir', type=str, default=LOGDIR,
+    parser.add_argument('--logdir', type=str, default=get_default_logdir(),
+                        help='Directory in which to store the logging '
+                        'information for TensorBoard.')
+    parser.add_argument('--checkpoint_dir', type=str, default=LOGDIR,
                         help='Directory in which to store the logging '
```
ibab: The help text should be adjusted. Maybe you want to say here that by default it is equal to the logdir.

ibab: Also, shouldn't this also use `get_default_logdir()`?

train.py
```diff
@@ -54,6 +59,45 @@ def get_arguments():
...
+        print("  Restoring...", end="")
+        saver.restore(sess, ckpt.model_checkpoint_path)
+        print(" Done.")
+        return global_step
+    else:
+        print(" No checkpoint found.")
+        return None
+
+
+def get_default_logdir():
```
ibab: Maybe it would be simpler to just call
```python
datestring = "{0:%Y-%m-%dT%H-%M-%S}".format(datetime.now())
default_logdir = os.path.join(LOGDIR, 'train', datestring)
```
somewhere close to the top of the script.
What do you think?

train.py
```diff
@@ -155,14 +200,28 @@ def main():
...
-        for step in range(args.num_steps):
+        saved_global_step = load(saver, sess, logdir)
+        if saved_global_step is None:
+            # The training step is start from saved_global_step + 1
+            # Therefore put -1 here if the new training starts.
+            saved_global_step = -1
+
+    except:
+        print("Something is wrong while restoring checkpoint. "
+              "We will terminate training to avoid accidental overwriting "
```
ibab: `accidental` -> `accidentally`

train.py
mecab: oops, somehow WIP contaminated... I'll fix it

ibab: I think by writing it this way, the user loses the ability to say "Place the logging output in a subdirectory of this directory", which was the original idea behind the `logdir` parameter.

mecab: Yes. So should I also add `--logdir_root` (I think it's more simple name than `--default_logdir_root`) to specify it?

ibab: That's a good idea 👍
How about adding a `--restore_from` option that specifies which model to start from?
(And starts a new subdirectory in the root logdir)

train.py
mecab: It's more simple and better. Thanks.

train.py
mecab: It's reasonable 👍

Maybe I should add some validations:
 - Reject specifying both of `--logdir` and `--logdir_root`
 - Reject specifying both of `--logdir` and `--restore_from`  
It is a bit danger a bit because it could unexpectedly overwrite previous model in `--logdir` by new training. Maybe we should add the interactive confirmation or a flag like `-f`, `--force` to ignore the guard, but I want leave it as another issue at this time.

ibab: Ah, I see. Yeah, that's a bit tricky.

mecab:
I believe I fixed the things.

While my English is not good enough to describe the complex path meanings, it's welcome to propose better descriptions. 🙇 🙇 🙇

ibab:
Looks good, thanks you!

PR #14, opened 2016-09-16
Readability refactor of wavenet.py

lemonzi:
Organize and document wavenet.py.

PR #58, opened 2016-09-21
Adjust network

ibab:
Multiple adjustments to the model definition:

1. Use separate 1x1 convolutions for residuals and skipped connections (suggested by @basveeling)
2. We can feed the raw waveform into the first convolutional layer instead of a one-hot encoding (as explained by @jyegerlehner)
3. Made the histogram summaries optional

I'm not super sure about 2 yet, it seems to produce roughly similar results to the one-hot encoding.
And it would probably be better to use quantized values instead of the original signal.

### DIFF
--- a/wavenet.py
+++ b/wavenet.py
@@ -18,13 +18,15 @@ class WaveNet(object):
                  dilations,
                  filter_width,
                  residual_channels,
-                 dilation_channels):
+                 dilation_channels,
+                 store_summaries=False):
         self.batch_size = batch_size
         self.channels = channels
         self.dilations = dilations
         self.filter_width = filter_width
         self.residual_channels = residual_channels
         self.dilation_channels = dilation_channels
+        self.store_summaries = store_summaries
 
 
     # A single causal convolution layer that can change the number of channels.
@@ -37,9 +39,10 @@ class WaveNet(object):
             return causal_conv(input_batch, weights_filter, 1)
 
 
-    def _create_dilation_layer(self, input_batch, layer_index, dilation, in_channels, dilation_channels):
-        '''Adds a single causal dilated convolution layer.'''
+    def _create_dilation_block(self, input_batch, layer_index, dilation, in_channels, dilation_channels):
+        '''Adds a causal dilated convolution with gated activation and 1x1 convolutions.'''
 
+        # Causal convolutions with gated activation function.
         weights_filter = tf.Variable(tf.truncated_normal(
             [self.filter_width, in_channels, dilation_channels],
             stddev=0.2,
@@ -53,16 +56,26 @@ class WaveNet(object):
 
         out = tf.tanh(conv_filter) * tf.sigmoid(conv_gate)
 
-        weights_dense = tf.Variable(tf.truncated_normal(
-            [1, dilation_channels, in_channels], stddev=0.2, name="dense"))
-        transformed = tf.nn.conv1d(out, weights_dense, stride=1,
-                                   padding="SAME", name="dense")
-        layer = 'layer{}'.format(layer_index)
-        tf.histogram_summary(layer + '_filter', weights_filter)
-        tf.histogram_summary(layer + '_gate', weights_gate)
-        tf.histogram_summary(layer + '_dense', weights_dense)
+        # The result of this 1x1 convolution is passed to the next block.
+        weights_residuals = tf.Variable(tf.truncated_normal(
+            [1, dilation_channels, in_channels], stddev=0.2, name="residuals"))
+        residuals = tf.nn.conv1d(out, weights_residuals, stride=1,
+                                 padding="SAME", name="residuals")
 
-        return transformed, input_batch + transformed
+        # The result of this 1x1 convolution is gathered at the end.
+        weights_skipped = tf.Variable(tf.truncated_normal(
+            [1, dilation_channels, in_channels], stddev=0.2, name="skipped"))
+        skipped = tf.nn.conv1d(out, weights_skipped, stride=1,
+                               padding="SAME", name="skipped")
+
+        if self.store_summaries:
+            layer = 'layer{}'.format(layer_index)
+            tf.histogram_summary(layer + '_filter', weights_filter)
+            tf.histogram_summary(layer + '_gate', weights_gate)
+            tf.histogram_summary(layer + '_residuals', weights_residuals)
+            tf.histogram_summary(layer + '_skipped', weights_skipped)
+
+        return skipped, input_batch + residuals
 
 
     def _preprocess(self, audio):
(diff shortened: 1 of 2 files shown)


wavenet.py
```diff
@@ -18,13 +18,15 @@ def __init__(self,
                  dilations,
                  filter_width,
                  residual_channels,
-                 dilation_channels):
+                 dilation_channels,
+                 store_summaries=False):
```
mecab: Should we add a flag to make `store_summaries` be changeable from `train.py`?

jyegerlehner:
Ah, I was working on a PR to do #1, that was branched from the add-biases PR, since I thought you suggested that ordering.

On #2, I tested my version of that code, and also didn't observe that it made any obvious difference, at least as things are now.

wavenet.py
```diff
@@ -53,16 +56,26 @@ def _create_dilation_layer(self, input_batch, layer_index, dilation, in_channels
...
+        # The result of this 1x1 convolution is passed to the next block.
+        weights_residuals = tf.Variable(tf.truncated_normal(
+            [1, dilation_channels, in_channels], stddev=0.2, name="residuals"))
+        residuals = tf.nn.conv1d(out, weights_residuals, stride=1,
+                                 padding="SAME", name="residuals")
 
-        return transformed, input_batch + transformed
+        # The result of this 1x1 convolution is gathered at the end.
+        weights_skipped = tf.Variable(tf.truncated_normal(
+            [1, dilation_channels, in_channels], stddev=0.2, name="skipped"))
```
jyegerlehner: I suggest adding a separate skip_channels param that serves as the number of output channels, instead of "in_channels" AKA residual_channels. I think we want the flexibility to have the skip contribution channels be different than the residual channels.

wavenet.py
```diff
@@ -108,14 +121,15 @@ def _create_network(self, input_batch):
             # Perform (+) -> ReLU -> 1x1 conv -> ReLU -> 1x1 conv to
             # postprocess the output.
             w1 = tf.Variable(tf.truncated_normal(
-                [1, self.residual_channels, int(self.channels / 2)], stddev=0.3,
+                [1, self.residual_channels, self.channels], stddev=0.3,
```
jyegerlehner: I would suggest making the input and output channels here equal to the new skip_channels param, if you went for what I was saying above.

Back on your thread for "generating good audio samples", one of the ideas I mentioned I'd like to try is setting the channels in these last two 1x1/ReLU layers to a larger number, and I think the skip_channels would cleanly fill that role.

wavenet.py
```diff
@@ -108,14 +121,15 @@ def _create_network(self, input_batch):
...
                 name="postprocess1"))
             w2 = tf.Variable(tf.truncated_normal(
-                [1, int(self.channels / 2), self.channels], stddev=0.3,
+                [1, self.channels, self.channels], stddev=0.3,
```
jyegerlehner: And then the input channels here would also be skip_channels.

wavenet.py
```diff
@@ -152,9 +166,10 @@ def predict_proba(self, waveform, name='wavenet'):
 
     def loss(self, input_batch, name='wavenet'):
         with tf.variable_scope(name):
-            input_batch = self._preprocess(input_batch)
-            encoded = self._one_hot(input_batch)
-            raw_output = self._create_network(encoded)
+            quantized = self._preprocess(input_batch)
+            encoded = self._one_hot(quantized)
+            inputs = tf.reshape(input_batch, [self.batch_size, -1, 1])
+            raw_output = self._create_network(inputs)
```
jyegerlehner: That looks familiar. :)

wavenet_params.json
```diff
@@ -2,8 +2,9 @@
     "filter_width": 2,
     "quantization_steps": 256,
     "sample_rate": 16000,
-    "dilations": [1, 2, 4, 8, 16, 32, 64, 128, 256,
-                  1, 2, 4, 8, 16, 32, 64, 128, 256],
+    "dilations": [1, 2, 4, 8, 16, 32, 64, 128, 256, 512,
+                  1, 2, 4, 8, 16, 32, 64, 128, 256, 512,
+                  1, 2, 4, 8, 16, 32, 64, 128, 256, 512],
     "residual_channels": 32,
     "dilation_channels": 16
```
jyegerlehner: "skip_channels": 1024 ? Pretty please?

jyegerlehner: Actually, of course it doesn't need to default to anything so big. In consideration for memory requirements.

jyegerlehner:
@ibab When I gave that list of things I'd like to try over on the "Generating good audio samples" I didn't expect you'd go code them all and put them in straight away :) Some of those ideas are a bit wild and thought I'd try them out and bring them back if they worked. I don't mind. Just a bit embarrassed that some of my nutty ideas are being shoe-horned straight into master.

wavenet.py
jyegerlehner: Wouldn't corresponding changes also need to be made in generate.py?

ibab:
As 2. hasn't really worked better in my overnight training run, I think I'll just close this PR and we can focus on getting @jyegerlehner's changes merged.

wavenet.py
lemonzi: Have we tested this to see if it helps with generation? I'd suggest moving this to a separate PR, as it's quite a big change in the network architecture.

wavenet.py
```diff
@@ -90,13 +103,13 @@ def _create_network(self, input_batch):
         outputs = []
         current_layer = input_batch
 
-        current_layer = self._create_causal_layer(current_layer, self.channels, self.residual_channels)
+        current_layer = self._create_causal_layer(current_layer, 1, self.residual_channels)
```
lemonzi: See L172.

wavenet.py
lemonzi: The intermediate tensor could be something else, but the number of hyperparameters is growing out of hand... How about skip_channels is a vector that defines the channels per layer, with optionally deeper convolutions than the 2 layers we have now? See https://github.com/ibab/tensorflow-wavenet/pull/14/files#diff-59f2c602fd3753c87268cb878555ae38R135 for an implementation.

lemonzi:
Oh, didn't see that, my bad. @jyegerlehner, what do you think about this skip_channels vector?

jyegerlehner:
@lemonzi  I like the idea in principle but am am bit torn as it is more complicated. Let's see if we can get opinions from ibab and others. Do you think it needs to be included in [PR 62](https://github.com/ibab/tensorflow-wavenet/pull/62) or would you prefer to bring it in as your own separate PR?

ibab:
@jyegerlehner @lemonzi: `skip_channels` is a good idea. Let's bring it in as a separate PR to make things easier to review.

I prefer `add_summaries` as well, after the reminder that it's kind of a TensorFlow convention.
I can extract the optional summaries part into a separate PR.

PR #52, opened 2016-09-20
Added tests for encoding and decoding functions

multivac61:

PR #60, opened 2016-09-21
cast output of decode to float32

woodshop:
The output of `decode` was `int32`, which caused `tf.audio_summary`
to throw an error in `generate.py`.


ibab:
I've added this change and a few others to the branch https://github.com/ibab/tensorflow-wavenet/tree/improve-mu-law some time ago and waited for #52.
As this fix is quite important, maybe I should just merge this PR now and create a PR from my branch.

PR #61, opened 2016-09-21
PEP8 Compatiblity / PEP8 test in CI

mecab:
In recent PRs,some reviews related to coding style can be seen.
I hope checking PEP8 compatibility in CI helps us.

I have also checked exist codes and fixed the errors / warnings.

https://travis-ci.org/mecab/tensorflow-wavenet/builds/161508727 here you can confirm the test will fail without compatibility.

Perhaps this PR should be merged after finishing big changes ongoing.

### DIFF
--- a/train.py
+++ b/train.py
@@ -28,6 +28,7 @@ WAVENET_PARAMS = './wavenet_params.json'
 STARTED_DATESTRING = "{0:%Y-%m-%dT%H-%M-%S}".format(datetime.now())
 SAMPLE_SIZE = 100000
 
+
 def get_arguments():
     parser = argparse.ArgumentParser(description='WaveNet example network')
     parser.add_argument('--batch_size', type=int, default=BATCH_SIZE,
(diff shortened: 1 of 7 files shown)


lemonzi:
Conflict hell, indeed... but I agree we need something like this.

ibab:
Yeah, we should try to keep the code quality high from now on, as fixing things up afterwards results in conflicts everywhere.

mecab:
Followed `master` at fc5417d

ibab:
I've looked through the changes again and couldn't find any problems, so I'm merging now.

ibab:
I, for one, welcome our new overlord `pep8`.

train.py
```diff
@@ -242,7 +247,8 @@ def main():
                 writer.add_summary(summary, step)
 
             duration = time.time() - start_time
-            print('step %d - loss = %.3f, (%.3f sec/step)' % (step, loss_value, duration))
+            print('step %d - loss = %.3f, (%.3f sec/step)'
```
mecab: I changed to use `.format()` instead of `%`, compared to the formatting in other lines.
But forgot to change format specifier...

I'll fix it soon.

PR #63, opened 2016-09-21
Slightly improve quantization and add test

ibab:
The current quantization procedure is slightly broken, as it introduces a small bias when encoding and decoding the signal.
I've modified it so that this doesn't happen and added a test.

### DIFF
--- a/test/test_mu_law.py
+++ b/test/test_mu_law.py
@@ -1,0 +1,35 @@
+import tensorflow as tf
+import numpy as np
+from wavenet import WaveNet
+
+class TestMuLaw(tf.test.TestCase):
+
+    def testEncodeDecode(self):
+
+        net = WaveNet(
+            batch_size=1,
+            channels=256,
+            dilations=[1],
+            filter_width=2,
+            residual_channels=256,
+            dilation_channels=256,
+        )
+
+        x = np.linspace(-1, 1, 1000).astype(np.float32)
+
+        # Test whether decoded signal is roughly equal to
+        # what was decoded before
+        with self.test_session() as sess:
+            x1 = sess.run(net.decode(net.encode(x)))
+
+        self.assertAllClose(x, x1, rtol=1e-1, atol=0.05)
+
+        # Make sure that re-encoding leaves the waveform invariant
+        with self.test_session() as sess:
+            x2 = sess.run(net.decode(net.encode(x1)))
+
+        self.assertAllClose(x1, x2)
+
+
+
+
--- a/test/test_mu_law.py
+++ b/test/test_mu_law.py
@@ -3,6 +3,7 @@ import tensorflow as tf
 
 from wavenet_ops import encode, decode
 
+
 class TestMuLaw(tf.test.TestCase):
 
     def testEncodeDecode(self):
--- a/test/test_mu_law.py
+++ b/test/test_mu_law.py
@@ -1,33 +1,25 @@
 import numpy as np
 import tensorflow as tf
 
-from wavenet import WaveNet
+from wavenet_ops import encode, decode
 
 class TestMuLaw(tf.test.TestCase):
 
     def testEncodeDecode(self):
 
-        net = WaveNet(
-            batch_size=1,
-            dilations=[1],
-            filter_width=2,
-            quantization_channels=256,
-            residual_channels=256,
-            dilation_channels=256,
-        )
-
         x = np.linspace(-1, 1, 1000).astype(np.float32)
+        channels = 256
 
         # Test whether decoded signal is roughly equal to
         # what was encoded before
         with self.test_session() as sess:
-            x1 = sess.run(net.decode(net.encode(x)))
+            x1 = sess.run(decode(encode(x, channels), channels))
 
         self.assertAllClose(x, x1, rtol=1e-1, atol=0.05)
 
         # Make sure that re-encoding leaves the waveform invariant
         with self.test_session() as sess:
-            x2 = sess.run(net.decode(net.encode(x1)))
+            x2 = sess.run(decode(encode(x1, channels), channels))
 
         self.assertAllClose(x1, x2)
 
--- a/test/test_mu_law.py
+++ b/test/test_mu_law.py
@@ -29,7 +29,3 @@ class TestMuLaw(tf.test.TestCase):
             x2 = sess.run(net.decode(net.encode(x1)))
 
         self.assertAllClose(x1, x2)
-
-
-
-
--- a/test/test_mu_law.py
+++ b/test/test_mu_law.py
@@ -1,7 +1,7 @@
 import numpy as np
 import tensorflow as tf
 
-from wavenet_ops import encode, decode
+from wavenet_ops import mu_law_encode, mu_law_decode
 
 
 class TestMuLaw(tf.test.TestCase):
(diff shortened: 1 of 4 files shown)


test/test_mu_law.py
```diff
@@ -0,0 +1,35 @@
+import tensorflow as tf
```
lemonzi: Should be:

```python
import numpy as np
import tensorflow as tf

from wavenet import WaveNet
```

Unless we treat wavenet as a separate package, in which case the empty line should be removed.

test/test_mu_law.py
```diff
@@ -0,0 +1,35 @@
...
+import numpy as np
+from wavenet import WaveNet
+
+class TestMuLaw(tf.test.TestCase):
+
+    def testEncodeDecode(self):
+
+        net = WaveNet(
+            batch_size=1,
+            channels=256,
```
lemonzi: `quantization_channels`, and as the last argument for clarity.

wavenet.py
```diff
@@ -65,25 +65,29 @@ def _create_dilation_layer(self, input_batch, layer_index, dilation, in_channels
         return transformed, input_batch + transformed
 
 
-    def _preprocess(self, audio):
+    def encode(self, audio):
         '''Quantizes waveform amplitudes.'''
-        with tf.name_scope('preprocessing'):
+        with tf.name_scope('encode'):
             mu = self.channels - 1
```
lemonzi: `self.quantization_steps` (will come with the upstream merge).

wavenet.py
```diff
@@ -65,25 +65,29 @@ def _create_dilation_layer(self, input_batch, layer_index, dilation, in_channels
...
 
     def decode(self, output):
-        mu = self.channels - 1
-        y = tf.cast(output, tf.float32)
-        y = 2 * (y / mu) - 1
-        x = tf.sign(y) * (1 / mu) * ((1 + mu)**abs(y) - 1)
-        return x
+        '''Recovers waveform from quantized values.'''
+        with tf.name_scope('decode'):
+            mu = self.channels - 1
```
lemonzi: `quantization_channels`

test/test_mu_law.py
```diff
@@ -0,0 +1,34 @@
...
+
+class TestMuLaw(tf.test.TestCase):
+
+    def testEncodeDecode(self):
+
+        net = WaveNet(
+            batch_size=1,
+            dilations=[1],
+            filter_width=2,
+            quantization_channels=256,
```
lemonzi: It should be the last argument

wavenet.py
```diff
@@ -106,22 +106,24 @@ def _create_dilation_layer(self,
...
-        x = tf.sign(y) * (1 / mu) * ((1 + mu)**abs(y) - 1)
-        return x
+        '''Recovers waveform from quantized values.'''
+        with tf.name_scope('decode'):
+            mu = self.quantization_channels - 1
+            # Map values back to [-1, 1]
+            casted = tf.cast(output, tf.float32)
+            y = 2 * (casted / mu) - 1
+            # Perform inverse of mu-law transformation
+            return tf.sign(y) * (1 / mu) * ((1 + mu)**abs(y) - 1)
```
mecab: Is it intentional no space? 
```
(1 + mu)**abs(y)
        ^^
```

ibab: I like putting no space around the exponentiation as an exception to the rule.
This is just a habit from writing a lot of scientific code.
What are people's opinions on this?

lemonzi: The line is also very long, maybe mimic `encode` with magnitude/signal?

lemonzi: Since we may add it to Travis... does the PEP8 linter complain?

lemonzi:
There should also be a test that checks that supplying an invalid `quanization_channels` (that is, <= 1) results in a predictable error, and possibly an assertion in the code that throws it.

wavenet.py
mecab: Yes, I understand putting no space around `**` is reasonable.
In terms of PEP8, it force putting space on some operators, but not for `**` (and I've just ran pep8) , so it's okay 👍

Let's go for it.

test/test_mu_law.py
```diff
@@ -0,0 +1,35 @@
+import numpy as np
+import tensorflow as tf
+
+from wavenet import WaveNet
+
```
mecab: Needs 2 spaces before class.

ibab:
How about `mu_law_encode` and `mu_law_decode`?

wavenet_ops.py
```diff
@@ -35,3 +35,26 @@ def causal_conv(value, filter_, dilation, name='causal_conv'):
                          [0, 0, 0],
                          [-1, tf.shape(value)[1], -1])
         return result
+
```
mecab: Only one line break is allowed before function.

(Sorry for pointing out only small things... 🙇)

ibab: I'm not sure what you mean, could you explain?

mecab: Well, maybe the line I commented has been moved after the following commit.

wavenet_ops.py
```diff
@@ -32,6 +34,29 @@ def causal_conv(value, filter_, dilation, name='causal_conv'):
             restored = tf.nn.conv1d(padded, filter_, stride=1, padding='SAME')
         # Remove excess elements at the end
         result = tf.slice(restored,
-                         [0, 0, 0],
-                         [-1, tf.shape(value)[1], -1])
+                          [0, 0, 0],
+                          [-1, tf.shape(value)[1], -1])
         return result
+
+
```
mecab: I was about to comment here.

> Only one line break is allowed before function.

mecab: Something is strange in Github.

I want to comment on `wavenet_ops.py` L42
```


def mu_law_encode(audio, quantization_channels):
```
should be 
```

def mu_law_encode(audio, quantization_channels):
```

test/test_mu_law.py
```diff
@@ -0,0 +1,30 @@
+import numpy as np
+import tensorflow as tf
+
+from wavenet_ops import mu_law_encode, mu_law_decode
+
+
+class TestMuLaw(tf.test.TestCase):
+
+    def testEncodeDecode(self):
```
lemonzi: Could you add the `len(np.unique(x)) == quantization_steps` assertion?

ibab: Is the code that I've added just now what you had in mind?

lemonzi: Yep, this is perfect, thanks!

wavenet_ops.py
ibab: Ah, I understand.
But shouldn't there be 2 lines there according to PEP8?

lemonzi: Yep, I think it's 1 line between methods or inner functions, but 2 lines for top-level functions and classes.

wavenet_ops.py
```diff
@@ -32,6 +34,29 @@ def causal_conv(value, filter_, dilation, name='causal_conv'):
...
+def mu_law_encode(audio, quantization_channels):
+    '''Quantizes waveform amplitudes.'''
+    with tf.name_scope('encode'):
+        mu = quantization_channels - 1
+        # Perform mu-law companding transformation (ITU-T, 1988).
+        magnitude = tf.log(1 + mu * tf.abs(audio)) / tf.log(1. + mu)
+        signal = tf.sign(audio) * magnitude
+        # Quantize signal to the specified number of levels
```
lemonzi: Add periods at the end of the comments?

lemonzi: Besides that (which we can ignore if you feel lazy about it), it LGTM. There's some docstrings missing, but we can add them later.

jyegerlehner:
LGTM. Yay, tests.

Issue #26, opened 2016-09-17
Add fast wavenet generation.


tomlepaine:
@ibab I noticed you saw the efficient wavenet generation implementation I wrote with my friends:

https://github.com/tomlepaine/fast-wavenet

Can we help you add it to `tensorflow-wavenet`?

ibab:
I was actually about to write you a mail :)
I'd be very happy to accept a PR if someone is interested in porting it over to this repo's implementation (Shouldn't be too difficult, as the core structure of the network is quite similar).
I should also be able to do it myself if necessary.

I noticed that your repo is licensed under GPL 3, which would be in conflict with this repo's MIT license if we were to copy code, though.

In the future I'd like to turn this repo into more of a community effort and move it to a Github organization, or the TensorFlow `models` repo.
Do you have any ideas or preferences as to where it should end up?

tomlepaine:
Great! I can start working on it tomorrow, unless you want to do it.

And I'm flexible on the license. I'll change it to MIT.

The TensorFlow `models` repo sounds good to me.

ibab:
Awesome! Feel free to comment here if you have a question or if there's something I can do.

tomlepaine:
Update: Just about finished with the port. Will clean up the code and run some speed tests tonight, then make a pull request.

tomlepaine:
Sorry for the delay!

It seems like the way we implemented our queue had a bug that made the states we pop incorrect a small percent of the time. We aren't sure of the exactly cause, but we think it is related to `tf.scatter_update`. We switched from our queue to `tf.FIFOQueue` and it seemed to fix the bug (and slightly improved our speed :tada:). Now (based on our tests) the fast generation exactly reproduces outputs and states of the slow generation.

We will make a pull request after lunch.

ibab:
Thanks for implementing this! The generation is now 🏃
I'm working on adding unit tests that test both the old and new ways of generating audio (and testing them against each other).

tomlepaine:
No problem. Glad it is working fast 😸  Let me know how the testing goes!

I ran a few manual tests before we added the skip connections. It seemed to match up. If you find any problems though, let me know.

PR #69, opened 2016-09-21
Fast generation

tomlepaine:
We added fast wavenet generation. Addresses issue #26.

**Verification**
- We compared the output of our fast generation with slow generation, and ensured it exactly matches.
- We also did some speed tests, and verify it is substantially faster 😸 

Any comments on style/formatting are welcome.

### DIFF
--- a/generate.py
+++ b/generate.py
@@ -69,7 +69,9 @@ def main():
 
     next_sample = net.predict_proba(samples)
     
-    sess.run(tf.initialize_all_variables())
+    if args.fast_generation:
+        sess.run(tf.initialize_all_variables())
+        sess.run(net.init_ops)
 
     variables_to_restore = {var.name[:-2]: var for var in tf.all_variables() if 'Variable' in var.name}
     saver = tf.train.Saver(variables_to_restore)
@@ -100,7 +102,7 @@ def main():
                 next_sample,
                 feed_dict={samples: window})
             
-        sample = np.random.choice(np.arange(quantization_steps), p=prediction)
+        sample = np.random.choice(np.arange(quantization_channels), p=prediction)
         waveform.append(sample)
         print('Sample {:3<d}/{:3<d}: {}'
               .format(step + 1, args.samples, sample))
--- a/generate.py
+++ b/generate.py
@@ -35,6 +35,8 @@ def get_arguments():
                         help='Path to output wav file')
     parser.add_argument('--save_every', type=int, default=SAVE_EVERY,
                         help='How many samples before saving in-progress wav')
+    parser.add_argument('--fast_generation', type=bool, default=True,
+                        help='Use fast generation')    
     return parser.parse_args()
 
 
@@ -60,13 +62,18 @@ def main():
         dilation_channels=wavenet_params['dilation_channels'],
         quantization_channels=wavenet_params['quantization_channels'],
         skip_channels=wavenet_params['skip_channels'],
-        use_biases=wavenet_params['use_biases'])
+        use_biases=wavenet_params['use_biases'],
+        fast_generation=args.fast_generation)
 
     samples = tf.placeholder(tf.int32)
 
     next_sample = net.predict_proba(samples)
+    
+    sess.run(tf.initialize_all_variables())
 
-    saver = tf.train.Saver()
+    variables_to_restore = {var.name[:-2]: var for var in tf.all_variables() if 'Variable' in var.name}
+    saver = tf.train.Saver(variables_to_restore)
+    
     print('Restoring model from {}'.format(args.checkpoint))
     saver.restore(sess, args.checkpoint)
 
@@ -75,15 +82,25 @@ def main():
     quantization_channels = wavenet_params['quantization_channels']
     waveform = np.random.randint(quantization_channels, size=(1,)).tolist()
     for step in range(args.samples):
-        if len(waveform) > args.window:
-            window = waveform[-args.window:]
+        if args.fast_generation:
+            window = waveform[-1]
+            outputs = [next_sample]
+            outputs.extend(net.push_ops)
+            outputs_list = sess.run(
+                outputs,
+                feed_dict={samples: window})
+            prediction = outputs_list[0]                                   
         else:
-            window = waveform
-        prediction = sess.run(
-            next_sample,
-            feed_dict={samples: window})
-        sample = np.random.choice(np.arange(quantization_channels),
-                                  p=prediction)
+            if len(waveform) > args.window:
+                window = waveform[-args.window:]
+            else:
+                window = waveform
+            outputs = [next_sample]
+            prediction = sess.run(
+                next_sample,
+                feed_dict={samples: window})
+            
+        sample = np.random.choice(np.arange(quantization_steps), p=prediction)
         waveform.append(sample)
         print('Sample {:3<d}/{:3<d}: {}'
               .format(step + 1, args.samples, sample))
--- a/generate.py
+++ b/generate.py
@@ -19,24 +19,45 @@ SAVE_EVERY = None
 
 def get_arguments():
     parser = argparse.ArgumentParser(description='WaveNet generation script')
-    parser.add_argument('checkpoint', type=str,
-                        help='Which model checkpoint to generate from')
-    parser.add_argument('--samples', type=int, default=SAMPLES,
-                        help='How many waveform samples to generate')
-    parser.add_argument('--logdir', type=str, default=LOGDIR,
-                        help='Directory in which to store the logging '
-                        'information for TensorBoard.')
-    parser.add_argument('--window', type=int, default=WINDOW,
-                        help='The number of past samples to take into '
-                        'account at each step')
-    parser.add_argument('--wavenet_params', type=str, default=WAVENET_PARAMS,
-                        help='JSON file with the network parameters')
-    parser.add_argument('--wav_out_path', type=str, default=None,
-                        help='Path to output wav file')
-    parser.add_argument('--save_every', type=int, default=SAVE_EVERY,
-                        help='How many samples before saving in-progress wav')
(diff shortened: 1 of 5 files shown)


lemonzi:
@ibab Since there are 3 authors, let's not squash it when merging.
@pkhorrami4, can you rebase the last 2 commits into one with a fixup?

wavenet_ops.py
```diff
@@ -34,4 +34,4 @@ def causal_conv(value, filter_, dilation, name='causal_conv'):
         result = tf.slice(restored,
                          [0, 0, 0],
                          [-1, tf.shape(value)[1], -1])
-        return result
+        return result
```
lemonzi: Missing newline.

wavenet.py
```diff
@@ -102,6 +106,50 @@ def _create_dilation_layer(self,
...
+
+    def _generator_causal_layer(self, input_batch, state_batch, in_channels, out_channels):
+        with tf.name_scope('causal_layer'):
+            weights_filter = tf.Variable(tf.truncated_normal(
+                [self.filter_width, in_channels, out_channels],
+                stddev=0.2,
+                name="filter"))
+
+            output = self._apply_weights(input_batch, state_batch, weights_filter)
+            print 'causal dtype:', output.dtype
```
lemonzi:
```python
print("Causal dtype: {}".format(output.dtype))
```

This looks like a debug statement though, maybe just remove it?

ibab:
Woot, this is awesome 🎉 
@lemonzi: Good point.

lemonzi:
BTW, thanks for spotting my mistake with `quantization_channels` in generate.py! I'll fix it on master so that we have a clean diff for this PR.

wavenet.py
pkhorrami4: You are right. It was a debug statement. It has been removed.

ibab:
@lemonzi: Yeah, we should try to extract as much as possible of the `generate.py` and `train.py` scripts into external functions and then unit-test everything.

tomlepaine:
@ibab Can we add a link in the `README` saying the repo used `fast-wavenet`?

ibab:
@tomlepaine: Yeah, that's a good idea.

ibab:
Merging #70 has introduced a few more conflicts, sorry for that.
But it makes sense to handle these as part of this PR, as fast generation with separate 1x1 convolutions is definitely something we want to be able to do.

generate.py
```diff
@@ -57,29 +59,47 @@ def main():
...
 
     next_sample = net.predict_proba(samples)
-
-    saver = tf.train.Saver()
+    
+    if args.fast_generation:
+        sess.run(tf.initialize_all_variables())
+        sess.run(net.init_ops)
+
+    variables_to_restore = {var.name[:-2]: var for var in tf.all_variables() if 'Variable' in var.name}
```
mecab: It looks just restoring all variables.
How about leave it as following?

```python
saver = tr.train.Saver()
```

generate.py
```diff
@@ -57,29 +59,47 @@ def main():
...
     decode = net.decode(samples)
 
-    quantization_steps = wavenet_params['quantization_steps']
-    waveform = np.random.randint(quantization_steps, size=(1,)).tolist()
+    quantization_channels = wavenet_params['quantization_channels']
+    waveform = np.random.randint(quantization_channels, size=(1,)).tolist()
     for step in range(args.samples):
-        if len(waveform) > args.window:
-            window = waveform[-args.window:]
+        if args.fast_generation:
```
mecab: I suggest simplify this line by extracting fast-generation-specific operation, like:

```python
if args.fast_generation:
    window = waveform[-1]
    outputs = [next_sample, net.push_ops]
else:
    window = waveform[-args.window:]
    outputs = [next_sample]
prediction = sess.run(outputs, feed_dict={samples: window})[0]
```

generate.py
```diff
@@ -57,29 +59,47 @@ def main():
...
+        fast_generation=args.fast_generation)
 
     samples = tf.placeholder(tf.int32)
 
     next_sample = net.predict_proba(samples)
-
-    saver = tf.train.Saver()
+    
+    if args.fast_generation:
+        sess.run(tf.initialize_all_variables())
```
mecab: `Saver` initializes the variables which was not restored, so this line is not necessary.
(@ibab let me know, though 😃: https://github.com/ibab/tensorflow-wavenet/issues/13#issuecomment-247645912)

tomlepaine: @mecab I'm not sure what you mean. Let me describe what we are trying to do and make sure we are on the same page. We create some new variables for the generation: `state_buffer` and `pointer` for each queue. We initialize them, and then initialize all the other variables from a checkpoint using the `saver` (see the code in your comment below).

I think this achieves that well enough. It could be cleaner though...

generate.py
tomlepaine: (See above)

generate.py
tomlepaine: Thanks, this is cleaner.

generate.py
```diff
@@ -57,29 +59,47 @@ def main():
...
-            window = waveform
-        prediction = sess.run(
-            next_sample,
-            feed_dict={samples: window})
-        sample = np.random.choice(np.arange(quantization_steps), p=prediction)
+            if len(waveform) > args.window:
+                window = waveform[-args.window:]
+            else:
+                window = waveform
+            outputs = [next_sample]
```
tomlepaine: Agreed. Your suggestion above fixes that.

wavenet.py
```diff
@@ -102,6 +106,49 @@ def _create_dilation_layer(self,
             tf.histogram_summary(layer + '_biases_dense', biases_dense)
 
         return transformed, input_batch + transformed
+    
+
+    def _apply_weights(self, input_batch, state_batch, weights):
+        weights_recurrent = weights[0, :, :]
```
tomlepaine: Fixed.

generate.py
ibab: Yeah, there are extra variables here that need to be initialized, so this line is necessary.

tomlepaine:
What do you think the best way to make a test for fast generation is? Make a new class `TestNetFastGeneration(tf.test.TestCase)`?

wavenet.py
```diff
@@ -102,6 +106,49 @@ def _create_dilation_layer(self,
...
+        weights_embedding = weights[1, :, :]
+
+        output = tf.matmul(input_batch, weights_embedding) + tf.matmul(state_batch,
```
tomlepaine: Will do.

lemonzi:
I'd rename the current class to `TestNet` and I would subclass it to `TestNetWithBiases`, `TestNetFastGeneration`, etc. overriding `SetUp`. We can always figure out a better way to do it, like giving it a list of dictionaries to initialize the net from and automatically testing each one.

tomlepaine:
Also what utility are using for PEP8 compatibility? autopep8, yapf?

generate.py
mecab: I now understand, thank you for the explanation.

mecab:
I'm just running `pep8` and fixing errors by hand...

mecab:
By the way, I ran your code and saw the following on training.
It is not happen in master. I guess some changes are needed.

```
Traceback (most recent call last):
  File "train.py", line 262, in <module>
    main()
  File "train.py", line 191, in main
    loss = net.loss(audio_batch)
  File "/Users/***/projects/tensorflow-wavenet/wavenet.py", line 341, in loss
    raw_output = self._create_network(encoded)
  File "/Users/***/projects/tensorflow-wavenet/wavenet.py", line 180, in _create_network
    self.residual_channels)
  File "/Users/***/projects/tensorflow-wavenet/wavenet.py", line 67, in _create_causal_layer
    return causal_conv(input_batch, weights_filter, 1)
  File "/Users/***/projects/tensorflow-wavenet/wavenet_ops.py", line 30, in causal_conv
    padded = tf.pad(value, padding)
  File "/Users/***/.virtualenvs/***/lib/python3.5/site-packages/tensorflow/python/ops/array_ops.py", line 1325, in pad
    return gen_array_ops._pad(tensor, paddings, name=name)

 (omit...)

  File "/Users/***/.virtualenvs/***/lib/python3.5/site-packages/tensorflow/python/framework/tensor_shape.py", line 641, in with_rank
    raise ValueError("Shape %s must have rank %d" % (self, rank))
ValueError: Shape (1, 256) must have rank 3
```

mecab:
In the last your commit the above is fixed.  Please ignore.

wavenet.py
```diff
@@ -23,7 +23,8 @@ def __init__(self,
                  dilation_channels,
                  skip_channels,
                  quantization_channels=2**8,
-                 use_biases=False):
+                 use_biases=False,
+                 fast_generation=True):
```
mecab: Consider make default `fast_generation=False` to avoid fail in the training.
(or pass `fast_generation=False` in train.py

tomlepaine:
Okay, I just realized something. Fast generation should only be used during generation. The tests only current test training...

So I think we will remove `TestNetFastGeneration` for now. After someone adds a generation test, we can add it back.

wavenet.py
tomlepaine: Sounds good. We'll do that.

lemonzi:
Sounds good! Tests for generation are actually one of the biggest
priorities right now.

El dj., 22 set. 2016 a les 15:23, Tom Le Paine (<notifications@github.com>)
va escriure:

tomlepaine:
Just about done I think. But we found a bug. most of the variables don't have names, but the skip weights do.

Our generator code currently assumes the skip weights don't have a name. We will fix that now.

tomlepaine:
Okay. We are ready to merge if you are. 😸

ibab:
LGTM. @lemonzi: Feel free to merge if you agree.

mecab:
LGTM, too. I'm generating samples by your code, and

Fast

PR #62, opened 2016-09-21
Add separate skip connection 1x1 convs

jyegerlehner:
This PR adds a separate 1x1 conv for the skip connection, as described by basveeling in [issue 48](https://github.com/ibab/tensorflow-wavenet/issues/48). ibab also an implementation of this change in [PR 58](https://github.com/ibab/tensorflow-wavenet/pull/58), the main difference being that his uses residual_channels for the skip connection channels, whereas this PR uses a separate config param "skip_channels" for the skip connections. Also, this PR is built on top of [PR #57](https://github.com/ibab/tensorflow-wavenet/pull/57), which adds biases.

PR #70, opened 2016-09-21
separate 1x1 conv for skip contributions

jyegerlehner:
Supercedes [PR62](https://github.com/ibab/tensorflow-wavenet/pull/62).

### DIFF
--- a/wavenet.py
+++ b/wavenet.py
@@ -20,6 +20,7 @@ class WaveNet(object):
                  residual_channels,
                  dilation_channels,
                  quantization_channels=2**8,
+                 skip_channels,
                  use_biases=False):
         '''Initializes the WaveNet model.
 
@@ -45,6 +46,7 @@ class WaveNet(object):
         self.dilation_channels = dilation_channels
         self.quantization_channels = quantization_channels
         self.use_biases = use_biases
+        self.skip_channels = skip_channels
 
     # A single causal convolution layer that can change the number of channels.
     def _create_causal_layer(self, input_batch, in_channels, out_channels):
@@ -62,7 +64,8 @@ class WaveNet(object):
                                layer_index,
                                dilation,
                                in_channels,
-                               dilation_channels):
+                               dilation_channels,
+                               skip_channels):
         '''Creates a single causal dilated convolution layer.'''
         weights_filter = tf.Variable(tf.truncated_normal(
             [self.filter_width, in_channels, dilation_channels],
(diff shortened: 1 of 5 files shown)


wavenet.py
```diff
@@ -171,7 +188,7 @@ def _create_network(self, input_batch):
 
             # We skip connections from the outputs of each layer, adding them
             # all up here.
-            total = sum(outputs)
+            total = tf.add_n(outputs)
```
ibab: Does this bring performance improvements?
The rationale behind using `sum` was that TensorFlow could perform the sum in parallel while calculating the layers, and didn't have to keep all skipped outputs around until the end.
Then again, the various sum operations might allocate a lot of memory.

wavenet.py
```diff
@@ -20,6 +20,7 @@ def __init__(self,
                  residual_channels,
                  dilation_channels,
                  quantization_channels=2**8,
+                 skip_channels,
```
ibab: The argument order needs to be changed here in order for the tests to pass.

wavenet.py
jyegerlehner: I thought tensorflow parallelizes all its ops for GPU. I was worried that the built-in python sum would end up creating tensors for each intermediate summation result. That also seems slower because, depending on how it gets implemented, has to write the intermediate tensor results to global memory multiple times.

But it really is an unrelated change to the intent of this PR so I guess it should be done separately if at all.

ibab:
Looks good to me!
Not sure what a good value for `skip_channels` would be, but 256 should do for now.

wavenet.py
```diff
@@ -88,20 +93,34 @@ def _create_dilation_layer(self,
...
+
+        # The 1x1 conv to produce the skip contribution.
+        weights_skip = tf.Variable(tf.truncated_normal(
+            [1, dilation_channels, skip_channels], stddev=0.01),
+                                   name="skip")
+        skip_contribution = tf.nn.conv1d(out,weights_skip,stride=1,
+            padding="SAME", name="skip")
+
         if self.use_biases:
             biases_dense = tf.Variable(tf.constant(0.0,shape=[in_channels]),
```
ibab: You could insert a space between `0.0,` and `shape` and a few lines below as well.

ibab:
I found 2 spots with missing spaces. I'll merge the PR once you add those.

wavenet.py
jyegerlehner: Anything for you, ibab!

Issue #68, opened 2016-09-21
Bootstrapping the generation with existing audio


ibab:
Currently, we start the generation with a randomly picked waveform sample.
I wonder what kind of effect that has, considering that the dilated convolutions won't be able to reach backwards beyond the beginning of the generated sample.
Maybe we should start off with one of the audio recordings.

mortont:
I've been testing different hyperparameters using this kind of generation, except I'm starting with the first half of a sample and attempting to generate the second half. So far I haven't had much more luck than the randomly generated waveform, but I've yet to adjust the `dilation_channels` and `residual_channels`. I can post the generated .wav file if there's interest.

ibab:
This will be fixed by #71.

lemonzi:
#71 has been merged, closing.

PR #71, opened 2016-09-22
Added audio seed for starting generator from a .wav

mortont:
Discussed in #68, this gives the option to input a .wav file as the waveform to start audio generation from. It will fall back to a single random seed if no file is provided.

### DIFF
--- a/generate.py
+++ b/generate.py
@@ -1,3 +1,5 @@
+from __future__ import division
+
 import argparse
 from datetime import datetime
 import json
@@ -8,7 +10,7 @@ import numpy as np
 import tensorflow as tf
 
 from wavenet import WaveNet
-from wavenet_ops import mu_law_decode
+from wavenet_ops import mu_law_decode, mu_law_encode
 
 SAMPLES = 16000
 LOGDIR = './logdir'
@@ -44,18 +46,13 @@ def write_wav(waveform, sample_rate, filename):
     librosa.output.write_wav(filename, y, sample_rate)
     print('Updated wav file at {}'.format(filename))
 
-def create_seed(filename, sample_rate, quantization_channels, percent=50):
-    audio = librosa.load(filename, sr=sample_rate, mono=True)
-    audio = audio[0]
+def create_seed(filename, sample_rate, quantization_channels, window_size=WINDOW):
+    audio, _ = librosa.load(filename, sr=sample_rate, mono=True)
 
-    mu = quantization_channels - 1
-    magnitude = np.log(1 + mu * np.abs(audio)) / np.log(1. + mu)
-    signal = np.sign(audio) * magnitude
-    quantized = (signal + 1) / 2 * mu
-    quantized = quantized.astype(np.int32)
+    quantized = mu_law_encode(audio, quantization_channels)
+    cut_index = tf.size(quantized) + tf.constant(window_size) - tf.constant(1)
 
-    cut_index = int(len(quantized) * (percent / 100))
-    return quantized[:cut_index].tolist()
+    return quantized[:cut_index]
 
 def main():
     args = get_arguments()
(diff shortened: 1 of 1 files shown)


generate.py
```diff
@@ -42,6 +44,18 @@ def write_wav(waveform, sample_rate, filename):
...
+    audio = librosa.load(filename, sr=sample_rate, mono=True)
+    audio = audio[0]
+
+    mu = quantization_channels - 1
+    magnitude = np.log(1 + mu * np.abs(audio)) / np.log(1. + mu)
+    signal = np.sign(audio) * magnitude
+    quantized = (signal + 1) / 2 * mu
+    quantized = quantized.astype(np.int32)
+
+    cut_index = int(len(quantized) * (percent / 100))
```
mecab: Maybe we need
```python
from __future__ import division
```
in the head of this file, as well as f6c79ee. Because `percent  / 100` can be zero on Python2.

generate.py
```diff
@@ -42,6 +44,18 @@ def write_wav(waveform, sample_rate, filename):
     librosa.output.write_wav(filename, y, sample_rate)
     print('Updated wav file at {}'.format(filename))
 
+def create_seed(filename, sample_rate, quantization_channels, percent=50):
+    audio = librosa.load(filename, sr=sample_rate, mono=True)
```
ibab: `audio, _ = ...` would be more elegant.

generate.py
```diff
@@ -42,6 +44,18 @@ def write_wav(waveform, sample_rate, filename):
...
+    audio = audio[0]
+
+    mu = quantization_channels - 1
```
ibab: Can you re-use the `encode` TensorFlow function defined in `wavenet_ops.py`?
It might be more convenient to return the cut audio waveform here, and then run
`sess.run(encode)` in the if statement below.

generate.py
```diff
@@ -42,6 +44,18 @@ def write_wav(waveform, sample_rate, filename):
     librosa.output.write_wav(filename, y, sample_rate)
     print('Updated wav file at {}'.format(filename))
 
+def create_seed(filename, sample_rate, quantization_channels, percent=50):
```
ibab: Maybe `args.window` would be a better indicator of how many samples to include in the seed, compared to `percent`.

generate.py
```diff
@@ -72,7 +86,10 @@ def main():
     decode = mu_law_decode(samples, wavenet_params['quantization_channels'])
 
     quantization_channels = wavenet_params['quantization_channels']
-    waveform = np.random.randint(quantization_channels, size=(1,)).tolist()
+    if args.wav_seed:
+        waveform = create_seed(args.wav_seed, wavenet_params['sample_rate'], quantization_channels)
```
ibab: This line seems quite long.
Consider breaking it into 2.

generate.py
mortont: Agreed! Thanks for he heads up, didn't even think of Python2 support.

generate.py
mortont: I agree, I think including a seed larger than `args.window` would be a bit superfluous anyway

generate.py
mortont: Sure, I'll change it to use `mu_law_encode`. Thanks for the tip!

mortont:
In general, would it be useful to send audio through the `trim_silence` function before slicing it into a seed?

mortont:
Changes have been pushed, let me know if there's anything else that needs changing.

ibab:
Looks great!
There now seems to a conflict with a previous change on `master`.
If you rebase on `master` and fix the conflict, we can merge the PR.

mortont:
Merged and fixed the conflict, should be good to go now!

ibab:
Thank you for the contribution!

mortont:
No problem! Thank you for everything you've done

lemonzi:
Closes #68.

PR #73, opened 2016-09-23
Avoid the crash with the audio which contains only silence

### LINKED ISSUE #59: Bug with trimming
I train the network with default hyper-parameters:

`Exception in thread Thread-2:
Traceback (most recent call last):
  File "/usr/lib/python2.7/threading.py", line 810, in __bootstrap_inner
    self.run()
  File "/usr/lib/python2.7/threading.py", line 763, in run
    self.__target(*self.__args, **self.__kwargs)
  File "/home/jesse/tensorflow_workspace/tensorflow-wavenet/audio_reader.py", line 87, in thread_main
    audio = trim_silence(audio[:, 0])
  File "/home/jesse/tensorflow_workspace/tensorflow-wavenet/audio_reader.py", line 47, in trim_silence
    return audio[indices[0]:indices[-1]]
IndexError: index 0 is out of bounds for axis 0 with size 0`

The error occurs at the last line of trim_silence method:

`return audio[indices[0]:indices[-1]]`

It seems that some audio consists of only silence and is blank after trimming.

mecab:
Also add warning output for the silence.

This will fix #59.

### DIFF
--- a/audio_reader.py
+++ b/audio_reader.py
@@ -23,7 +23,7 @@ def load_generic_audio(directory, sample_rate):
     for filename in files:
         audio, _ = librosa.load(filename, sr=sample_rate, mono=True)
         audio = audio.reshape(-1, 1)
-        yield audio
+        yield audio, filename
 
 
 def load_vctk_audio(directory, sample_rate):
@@ -44,7 +44,9 @@ def trim_silence(audio, threshold=0.3):
     energy = librosa.feature.rmse(audio)
     frames = np.nonzero(energy > threshold)
     indices = librosa.core.frames_to_samples(frames)[1]
-    return audio[indices[0]:indices[-1]]
+
+    # Note: indices can be an empty array, if the whole audio was silence.
+    return audio[indices[0]:indices[-1]] if indices.size else audio[0:0]
 
 
 class AudioReader(object):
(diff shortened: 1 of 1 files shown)


jyegerlehner:
Hi mecab,
I get the following exception when I run this:
```
Exception in thread Thread-2:
Traceback (most recent call last):
  File "/usr/lib/python2.7/threading.py", line 810, in __bootstrap_inner
    self.run()
  File "/usr/lib/python2.7/threading.py", line 763, in run
    self.__target(*self.__args, **self.__kwargs)
  File "/home/jd/dev/models/tensorflow/tensorflow-wavenet/audio_reader.py", line 94, in thread_main
    audio = trim_silence(audio[:, 0], filename)
  File "/home/jd/dev/models/tensorflow/tensorflow-wavenet/audio_reader.py", line 48, in trim_silence
    if indices:
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()

```

Yes, I'm one of those barbarians who runs python 2.7.

mecab:
@jyegerlehner 
Thanks for pointing it out, i'll fix it.
(It happen also in python3... I pushed wrong branch.)

mecab:
fixed.

lemonzi:
I would print the warning in `AudioReader`; `trim_silence` doesn't need to know about filenames.

audio_reader.py
```diff
@@ -39,12 +39,19 @@ def load_vctk_audio(directory, sample_rate):
         yield audio, speaker_id
 
 
-def trim_silence(audio, threshold=0.3):
+def trim_silence(audio, filename, threshold=0.3):
```
ibab: It was kind of nice that the `trim_silence` function  could be used on arbitrary waveforms, without supplying a filename.
We can keep the check for an empty shape, and create the warning message at the point below where `trim_silence`, then the file name wouldn't be needed here.

audio_reader.py
```diff
@@ -78,13 +85,13 @@ def thread_main(self, sess):
...
             iterator = load_generic_audio(self.audio_dir, self.sample_rate)
-            for audio in iterator:
+            for audio, filename in iterator:
                 if self.coord.should_stop():
                     self.stop_threads()
                     stop = True
                     break
                 # Remove silence
-                audio = trim_silence(audio[:, 0])
+                audio = trim_silence(audio[:, 0], filename)
```
ibab: We can replace this with something like
```python
audio = trim_silence(audio[:, 0])
if not audio:
    # Print warning message
```

ibab:
Thanks for fixing this!
I've made some comments on how you could implement @lemonzi's suggestion.

mecab:
Yeah, I like it, too. I've fixed it.

ibab:
Nice work! Merging this now.

PR #74, opened 2016-09-23
Save checkpoint on exit

jyegerlehner:


train.py
```diff
@@ -224,7 +226,9 @@ def main():
     reader.start_threads(sess)
 
     try:
+        last_save_step = None
         for step in range(saved_global_step + 1, args.num_steps):
+
```
jyegerlehner: Sigh

train.py
```diff
@@ -224,7 +226,9 @@ def main():
     reader.start_threads(sess)
 
     try:
+        last_save_step = None
```
mecab: Maybe `last_save_step = saved_global_step` is more stable, considering training on restored checkpoint.

jyegerlehner: Thx

train.py
```diff
@@ -224,7 +226,9 @@ def main():
     reader.start_threads(sess)
 
     try:
+        last_save_step = saved_global_step
```
lemonzi: `last_saved_step`?

train.py
```diff
@@ -252,8 +256,12 @@ def main():
 
             if step % 50 == 0:
                 save(saver, sess, logdir, step)
+                last_save_step = step
 
     finally:
+        if step > last_save_step:
+            print()
```
lemonzi: What does this `print` do? Could you add a comment?

train.py
jyegerlehner: Whoopsie daisy. Thought I ran that. Thx

train.py
jyegerlehner: When you hit ctrl-C to kill it, a ^C is displayed. The print() puts the saving message on the next line after the ^C. Easier to read.

lemonzi: I see, that makes a lot of sense. In that case, could you add an `except KeyboardInterrupt:` block before `finally`, and do it there together with an explanatory comment (`# Introduce a line break after ^C`, or something like that)? It would make it much clearer, and it's a very neat trick; having it well documented will help other people (like me) learn it and use it elsewhere.

This is only relevant, though, if you are displaying stdout on the terminal, as opposed to redirecting and writing to a log file.

I'm also wondering if there's a more explicit way of printing an empty line. Maybe `print(end='\n')`? It feels like too much though, I think the current implementation is cleaner.

jyegerlehner: OK fellers I think I've addressed the comments so far. And squashed.

lemonzi: Great, thanks! For next time, GitHub squashes automatically when merging and gives that cool "commited with" message rather than creating a merge commit, so no need for that 😄

jyegerlehner: Ah, thanks. What will they think of next.

Issue #65, opened 2016-09-21
add regularization, dropout and batch norm?


r-zemblys:
Has anybody got loss lower than ~2? Tried couple of configurations (default, 3 and 4 stacks of 10 dilation layers), but loss does not get lower, suggesting the network is not learning anymore. 

Also, there is what happened happened after ~30k steps:
![training](https://cloud.githubusercontent.com/assets/8914323/18709029/99d32c48-8006-11e6-8a33-ecd8388afde7.png)
I believe this is the same problem as reported in [#30](https://github.com/ibab/tensorflow-wavenet/issues/30). There is what happens with weights:
![weights](https://cloud.githubusercontent.com/assets/8914323/18709214/abe1ffbc-8007-11e6-9454-fabca3cfc3e5.png)

Now running the same network with l2 norm regularization added.

And one more note: training just stops after 44256 steps (already happened twice) without any warnings or errors, despite of num_steps=50000

dnuffer:
I've observed the same things. I looked at the code to see what might be hanging and didn't find any red flags. I thought the hang might be related to my setup: CUDA 8.0rc (required for Pascal support), cuDNN 5.1, and tensorflow built from source (git master from 9/20)

ibab:
The hanging is probably caused by the background audio processing crashing. (Especially if if the CPU/GPU are idle once it stops).
Usually, there should be a backtrace that can help us find the reason it crashed.
Which commit did you observe the problem with?
There was a bug where we simply stop processing audio once we've seen every file once.
It might be that you're on an older commit that had this problem.

I've been trying to find a solution to the gradient jumping to large values at large step numbers, but don't have any amazing solutions at the moment.
It seems to be related to the ReLU activations in the last few layers of the network.
I've tried clipping the gradients, which didn't have an effect on this problem.
Replacing the ReLU activations with Tanh seems to fix it completely, but the network doesn't converge quite as quickly as with ReLU.

lelayf:
@ibab  I'm experiencing the stalling with the latest commit. 
@r-zemblys if you resume training at the checkpoint right before gradients implosion with a lower learning rate does it still behave the same ?

r-zemblys:
@lelayf i've used learning rate of 0.01 to get that loss curve above. Train saver only  stores last  5 checkpoints so I'm not able to try lowering learning rate right before gradient implosion. 

@ibab I was indeed using older commit. Latest one does not have stalling problem. 

[...]

PR #75, opened 2016-09-23
Added L2 norm regularization term to the loss to prevent gradient exp…

r-zemblys:
Fixes [#65](https://github.com/ibab/tensorflow-wavenet/issues/65) issue with gradient explosion

### DIFF
--- a/wavenet.py
+++ b/wavenet.py
@@ -211,7 +211,7 @@ class WaveNet(object):
             return tf.reshape(last, [-1])
 
 
-    def loss(self, input_batch, name='wavenet'):
+    def loss(self, input_batch, C=5e-4, name='wavenet'):
         '''Creates a WaveNet network and returns the autoencoding loss.
 
         The variables are all scoped to the given name.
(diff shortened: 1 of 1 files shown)


wavenet.py
```diff
@@ -235,7 +235,16 @@ def loss(self, input_batch, name='wavenet'):
...
+                                    for v in tf.trainable_variables()])
+                
+                # Add the regularization term to the loss
+                total_loss=reduced_loss + C * l2_loss
 
                 tf.scalar_summary('loss', reduced_loss)
+                tf.scalar_summary('l2_loss', l2_loss)
+                tf.scalar_summary('total_loss', total_loss)
 
         return reduced_loss
```
jyegerlehner: Wouldn't we want to be returning total_loss here?

wavenet.py
```diff
@@ -211,7 +211,7 @@ def predict_proba(self, waveform, name='wavenet'):
             return tf.reshape(last, [-1])
 
 
-    def loss(self, input_batch, name='wavenet'):
+    def loss(self, input_batch, C=5e-4, name='wavenet'):
```
jyegerlehner: C looks like something we would want to configure from command line or wavenet_params.json.

jyegerlehner: ...and I might rather have that default to zero.

wavenet.py
r-zemblys: Right, sorry about that

wavenet.py
r-zemblys: I would add it as a command line parameter. But defaulting it to zero is a good idea

r-zemblys:
I'm closing this PR as I messed up with my branch

PR #76, opened 2016-09-23
L2 norm regularization term to the loss to prevent gradient exp…

r-zemblys:
Reopened this instead of [#75](https://github.com/ibab/tensorflow-wavenet/pull/75).
FIX: return total loss. Thanks @jyegerlehner 
FIX: default C is now 0. Thanks @jyegerlehner 
TODO: @jyegerlehner : C looks like something we would want to configure from command line or wavenet_params.json.

### DIFF
--- a/train.py
+++ b/train.py
@@ -206,7 +206,7 @@ def main():
     sess = tf.Session(config=tf.ConfigProto(log_device_placement=False))
     init = tf.initialize_all_variables()
     sess.run(init)
-    stop
+    
     # Saver for storing checkpoints of the model.
     saver = tf.train.Saver()
 
--- a/train.py
+++ b/train.py
@@ -27,7 +27,7 @@ LEARNING_RATE = 0.02
 WAVENET_PARAMS = './wavenet_params.json'
 STARTED_DATESTRING = "{0:%Y-%m-%dT%H-%M-%S}".format(datetime.now())
 SAMPLE_SIZE = 100000
-L2_REGULARIZATION_STRENGTH = None
+L2_REGULARIZATION_STRENGTH = 0.
 
 def get_arguments():
     parser = argparse.ArgumentParser(description='WaveNet example network')
@@ -191,7 +191,7 @@ def main():
         skip_channels=wavenet_params["skip_channels"],
         quantization_channels=wavenet_params["quantization_channels"],
         use_biases=wavenet_params["use_biases"])
-    loss = net.loss(audio_batch, C=args.l2_regularization_strength)
+    loss = net.loss(audio_batch, l2_regularization_strength=args.l2_regularization_strength)
     optimizer = tf.train.AdamOptimizer(learning_rate=args.learning_rate)
     trainable = tf.trainable_variables()
     optim = optimizer.minimize(loss, var_list=trainable)
--- a/train.py
+++ b/train.py
@@ -27,7 +27,7 @@ LEARNING_RATE = 0.02
 WAVENET_PARAMS = './wavenet_params.json'
 STARTED_DATESTRING = "{0:%Y-%m-%dT%H-%M-%S}".format(datetime.now())
 SAMPLE_SIZE = 100000
-
+L2_REGULARIZATION_STRENGTH=0.
 
 def get_arguments():
     parser = argparse.ArgumentParser(description='WaveNet example network')
@@ -64,6 +64,9 @@ def get_arguments():
     parser.add_argument('--sample_size', type=int, default=SAMPLE_SIZE,
                         help='Concatenate and cut audio samples to this many '
                         'samples')
+    parser.add_argument('--l2_regularization_strength', type=float, 
+                        default=L2_REGULARIZATION_STRENGTH,
+                        help='L2 regularization strength ')
     return parser.parse_args()
 
 
@@ -188,7 +191,7 @@ def main():
         skip_channels=wavenet_params["skip_channels"],
         quantization_channels=wavenet_params["quantization_channels"],
         use_biases=wavenet_params["use_biases"])
-    loss = net.loss(audio_batch)
+    loss = net.loss(audio_batch, C=args.l2_regularization_strength)
     optimizer = tf.train.AdamOptimizer(learning_rate=args.learning_rate)
     trainable = tf.trainable_variables()
     optim = optimizer.minimize(loss, var_list=trainable)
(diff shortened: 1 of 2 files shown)


lemonzi:
The biases shouldn't be regularized: http://stats.stackexchange.com/questions/86991/reason-for-not-shrinking-the-bias-intercept-term-in-regression.

wavenet.py
```diff
@@ -210,7 +210,7 @@ def predict_proba(self, waveform, name='wavenet'):
                             [1, self.quantization_channels])
             return tf.reshape(last, [-1])
 
-    def loss(self, input_batch, name='wavenet'):
+    def loss(self, input_batch, C=0., name='wavenet'):
```
ibab: The `C` parameter could use a more descriptive name, like `regularization_strength`, `regularization_coefficient`, or just `regularization`.
We should also set it to `None` by default, and not create any of the regularization ops below if it's `None`.
That way we the user doesn't anything to the TensorFlow graph when they don't want to use regularization.

train.py
```diff
@@ -27,7 +27,7 @@
 WAVENET_PARAMS = './wavenet_params.json'
 STARTED_DATESTRING = "{0:%Y-%m-%dT%H-%M-%S}".format(datetime.now())
 SAMPLE_SIZE = 100000
-
+L2_REGULARIZATION_STRENGTH=0.
```
ibab: It would be nice to put spaces around the `=` and to set the parameter equal to `None` by default.

wavenet.py
```diff
@@ -234,7 +234,16 @@ def loss(self, input_batch, name='wavenet'):
                     prediction,
                     tf.reshape(shifted, [-1, self.quantization_channels]))
                 reduced_loss = tf.reduce_mean(loss)
-
+
```
ibab: There are some extra spaces in this line that can be removed.

wavenet.py
```diff
@@ -234,7 +234,16 @@ def loss(self, input_batch, name='wavenet'):
                     prediction,
                     tf.reshape(shifted, [-1, self.quantization_channels]))
                 reduced_loss = tf.reduce_mean(loss)
-
+                
+                # L2 regularization for all trainable parameters
+                l2_loss = tf.add_n([tf.nn.l2_loss(v) 
+                                    for v in tf.trainable_variables()])
+
```
ibab: extra spaces in this line.

wavenet.py
```diff
@@ -234,7 +234,16 @@ def loss(self, input_batch, name='wavenet'):
...
                 reduced_loss = tf.reduce_mean(loss)
-
+                
+                # L2 regularization for all trainable parameters
+                l2_loss = tf.add_n([tf.nn.l2_loss(v) 
+                                    for v in tf.trainable_variables()])
+                                        
+                # Add the regularization term to the loss
+                total_loss = reduced_loss + C * l2_loss
+
```
ibab: extra spaces in this line.

wavenet.py
```diff
@@ -234,7 +234,17 @@ def loss(self, input_batch, name='wavenet'):
                     prediction,
                     tf.reshape(shifted, [-1, self.quantization_channels]))
                 reduced_loss = tf.reduce_mean(loss)
-
+                
+                # L2 regularization for all trainable parameters
+                l2_loss = tf.add_n([tf.nn.l2_loss(v) 
+                                    for v in tf.trainable_variables() 
+                                    if not('bias' in v.name)])
+
```
ibab: extra spaces here.

wavenet.py
```diff
@@ -234,7 +234,17 @@ def loss(self, input_batch, name='wavenet'):
...
-
+                
+                # L2 regularization for all trainable parameters
+                l2_loss = tf.add_n([tf.nn.l2_loss(v) 
+                                    for v in tf.trainable_variables() 
+                                    if not('bias' in v.name)])
+                
+                # Add the regularization term to the loss
+                total_loss = reduced_loss + C * l2_loss
+
```
ibab: extra spaces here.

train.py
```diff
@@ -203,7 +206,7 @@ def main():
     sess = tf.Session(config=tf.ConfigProto(log_device_placement=False))
     init = tf.initialize_all_variables()
     sess.run(init)
-
+
```
lemonzi: Trailing space.

train.py
lemonzi: Should this be part of `wavenet_params.json`? It doesn't change how generation works, but it will result in very different models.

ibab: I can imagine that users might want to stop the training and restart it with a different regularization parameter. Although this would also work by editing `wavenet_params.json`, I think that having it as part of the training script makes things easier.
It also seems natural to have it next to the learning rate.

lemonzi: Then, it might make sense to have a second JSON file with training parameters: L2 regularization, learning rate, learning rate decay, dropout rate, ...

lemonzi: Oh, and it should also be easy to automatically add a flag for each JSON field allowing manual overrides.

ibab: I like the idea of a second JSON file.
Maybe there are already some Python libraries out there that handle this for us?

lemonzi: The easy way is to define everything using flags and argparse, and then supply besides the flags a list of JSON files. The JSON files will be sequentially read and merged with the current set of options, and finally the dictionary from argparse will be merged with the current set, overriding everything. I can try and make a PR with this. I don't know of any libraries that may make this easier. The question then is whether the flags are global or we allow nesting via a magic nesting character in the flag name like `/` or `.`.

We could also define flags at different places (each file defines the global flag it needs), which works well with large codebases, but maybe the repo is not large enough for this.

lemonzi: I'll open an issue for this, for the time being the regularization can stay as a separate flag.

jyegerlehner: > I like the idea of a second JSON file.

I dislike the idea of a second JSON file.

> a second JSON file with training parameters

Weren't you the one complaining about too many hyperparameters? Well I'm going to complain about too many config files. How about just putting two objects within a single file, one for model parameters and a second for training params.

ibab: The reason I was thinking of splitting them up is that we're currently using the `wavenet_params.json` file to encode the dimensions of the network. The file needs to stay the same between training and generation.
That's why I'm a bit reluctant to put training parameters in there as well, but I agree that we need a place to put the parameters.
Rest assured that we'll try to find a sane solution, which may also just involve putting everything into a single file. The discussion on this is in #79 and #80.

ibab:
You've changed the `None` default back to `0`.
Was there a problem with using `None`?

Also, there's a conflict with `train.py`.
If you want, I can take your changes and apply them to the project (your name will appear on the commit).

r-zemblys:
There was a problem setting `None` as default in `argparse`, so I just came back to 0. 
Yes please, resolve that conflict.

ibab:
I've added your changes to #88, so they will end up on `master` once it's merged. Thank you!

PR #78, opened 2016-09-23
Formatting fixes.

lemonzi:


train.py
```diff
@@ -43,7 +43,7 @@ def get_arguments():
                         help='Directory in which to store the logging '
                         'information for TensorBoard. '
                         'If the model already exists, it will restore '
-                        'the state and will continue training. '
+                        'the state and will continue training.'
```
ibab: Won't this result in no space between the `.` and the `Cannot` that will get added to the line?

lemonzi: Yes, my bad... I checked the other cases where this happens, but didn't see this one. Sorry about that.

Issue #42, opened 2016-09-19
Initialization of variables


jyegerlehner:
Are the magic numbers for weight initialization chosen for particular reasons (e.g. stddev=0.3)? Would we prefer something like tf.contrib.layers.xavier_initializer?

ibab:
There's no reason behind picking `stddev=0.3` at the moment, and using Xavier initialization would be nice. Note that the learnable filter isn't just an `(in_channels, out_channels)` matrix, as needed for the Xavier initializer, but an `(filter_width, in_channels, out_channels)` tensor.
Maybe stacking together `filter_width` Xavier initializers would work well.

jyegerlehner:
Right,well there's also tf.contrib.layers.xavier_initializer_conv2d. But no tf.contrib.layers.xavier_initializer_conv1d. But the only difference is amonst these is how they compute fan-in and fan-out; after that they compute the normal-std-dev or uniform-width from that. Should be straightforward to roll our own.

ibab:
@jyegerlehner: I didn't actually know about `xavier_initializer_conv2d`. Looking at the code, I can see that it only accesses the last 2 dimensions of the tensor, so we can just use it for our rank 3 tensors.
This example just works:
```python
import tensorflow as tf

initializer = tf.contrib.layers.xavier_initializer_conv2d()
variable = tf.get_variable('x', shape=[2, 32, 32], initializer=initializer)

sess = tf.Session()
sess.run(tf.initialize_all_variables())
print(sess.run(variable))
```

jyegerlehner:
OK I'll try to get to this in the not-too-distant future. If anyone else wants to pick it up please speak up so we don't end up duplicating effort.

ibab:
@jyegerlehner: It would make sense to define a `_create_variable(name, shape)` method that defines a new variable and initializes it using the Xavier initializer, so that it becomes easier to make the initializer configurable.

jyegerlehner:
@ibab Sounds like a good idea.

PR #82, opened 2016-09-24
Xavier initialization

jyegerlehner:
As discussed [here](https://github.com/ibab/tensorflow-wavenet/issues/42).

Regarding the failing test:  I guess the problem is that we switched from tf.Variable() to tf.get_variable(), and the self.net member variable is being re-used within the tf.test.TestCase.test_session. The second test tries to create a graph with all the same variable names. 
```
ValueError: Variable wavenet/causal_layer/filter already exists, disallowed. Did you mean to set reuse=True in VarScope? 
```
I tried for hours to refactor that with no success. I'm a C++ person, OK? I just flounder around in python. It usually works out. Any chance one of you python people could point me in the right direction please?

### DIFF
--- a/wavenet.py
+++ b/wavenet.py
@@ -1,6 +1,7 @@
 import tensorflow as tf
 
-from wavenet_ops import causal_conv, mu_law_encode
+from wavenet_ops import causal_conv, mu_law_encode, create_variable, \
+                        create_bias_variable
 
 
 class WaveNet(object):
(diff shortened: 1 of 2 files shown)


ibab:
`tf.get_variable` seems to be the recommended way of using the initializers, but it's not strictly necessary to use it. You can do something like
```
import tensorflow as tf

initializer = tf.contrib.layers.xavier_initializer_conv2d()
variable = tf.Variable(initializer(shape=[2, 32, 32]), name='x')

sess = tf.Session()
sess.run(tf.initialize_all_variables())
print(sess.run(variable))
```

The way we currently create variables is a bit messy when it comes to instantiating `WaveNet` multiple times, but it should do for now.

jyegerlehner:
OK thanks ibab. I just didn't want to disappoint you by retreating from your suggested tf.get_variable() approach.

ibab:
No worries :)
I was thinking about `tf.get_variable` because it could help us construct both the loss and the generation function in the same script. Currently, that's not possible as they will create new variables each time.
But it's something we can solve at a later point.

ibab:
I feel like the `create_variable` and `create_bias` functions might be better placed in `wavenet.py`, as they are specific to how we create the model in that file.
But we can have them in `wavenet_ops.py` for now and think about it later when we reorganize things a bit.

PR #84, opened 2016-09-25
Added tests to the test_mu_law.py file

multivac61:
This is a PR containing new tests added to `test_mu_law.py`

Due to a mess up when rebasing my fork, [this PR](https://github.com/ibab/tensorflow-wavenet/pull/52) contains the same code and should be removed after pulling this PR

### DIFF
--- a/test/test_mu_law.py
+++ b/test/test_mu_law.py
@@ -4,6 +4,32 @@ import tensorflow as tf
 from wavenet_ops import mu_law_encode, mu_law_decode
 
 
+# A set of mu law encode/decode functions implemented
+# in numpy
+def manual_mu_law_encode(signal, quantization_channels):
+    # Manual mu-law companding and mu-bits quantization
+    mu = quantization_channels - 1
+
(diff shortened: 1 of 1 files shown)


ibab:
Thanks for taking the time to rebase your pull request!

I've noticed that the tests are quite explicitly testing the encode/decode functionality against the same formulas coded in numpy instead of TensorFlow.
This means that the tests will fail if we only make a tiny adjustment to the encoding/decoding formulas.
In general, it's good practice to write unit tests that check abstract properties of the code instead of its explicit form, as that gives us some room to adjust it in useful ways without breaking the tests.

As we don't plan to change the encoding/decoding formulas, I'll merge your pull request, but this is something to keep in mind for the future :)

multivac61:
Aah okay. By abstract properties do you mean more like correct shapes of outputs, data types etc?

ibab:
@multivac61 Yeah, checking the shapes and data types is good.
Also, making sure that decoding in some sense undoes the encoding (ignoring quantization errors).
I think both functions are covered very well now.

PR #88, opened 2016-09-25
Restructure code and test generation

ibab:
I've rewritten some of the code today to make it easier to test and extend in the future.
Among other things, I've
 - Moved the code files into a `wavenet` package
 - Defined all variables in `Wavenet` up front so that they can be shared between training and generation methods
 - Renamed `WaveNet` to `WaveNetModel` because I want to introduce another class `WaveNet` that manages all the TensorFlow stuff around `WaveNetModel` so that the `train.py` and `generate.py` scripts should become very simple.
 - Added tests for generation (including a passing test that compares the old and new generation methods)
 - Fixed a bug in how we handled the boolean argument in `generate.py`

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,7 +106,7 @@ python generate.py --wav_out_path=generated.wav --save_every 2000 --samples 1600
 ```
 
 Fast generation is enabled by default.
-It uses the implementation from the [Fast Wavenet](https://github.com/tomlepaine/fast-wavenet) repository. You can follow the link for an explanation of how it works. This reduces the time needed to generate samples to a few minutes.
+It uses the implementation from the [Fast Wavenet](https://github.com/tomlepaine/fast-wavenet) repository.
 You can follow the link for an explanation of how it works.
 This reduces the time needed to generate samples to a few minutes.
 
--- a/README.md
+++ b/README.md
@@ -105,7 +105,10 @@ Passing `--save_every` in addition to `--wav_out_path` will save the in-progress
 python generate.py --wav_out_path=generated.wav --save_every 2000 --samples 16000 model.ckpt-1000
 ```
 
-Fast generation is enabled by default. It uses the implementation from the [Fast Wavenet](https://github.com/tomlepaine/fast-wavenet) repository. You can follow the link for an explanation of how it works. This reduces the time needed to generate samples to a few minutes.
+Fast generation is enabled by default.
+It uses the implementation from the [Fast Wavenet](https://github.com/tomlepaine/fast-wavenet) repository. You can follow the link for an explanation of how it works. This reduces the time needed to generate samples to a few minutes.
+You can follow the link for an explanation of how it works.
+This reduces the time needed to generate samples to a few minutes.
 
 To disable fast generation:
 ```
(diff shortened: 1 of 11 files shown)


jyegerlehner:
>  	Add bias support to fast generation

Holy crap that was fast.

jyegerlehner:
@ibab, I pulled this branch and trained with it. Error dropped rapidly all the way to zero. Then I generated from it and got white noise. I think the only way error could drop to zero is if the net had access to future values, i.e. not causal.

ibab:
@jyegerlehner: Okay, thanks for finding that out. I'll start looking for bugs.

ibab:
Yep, it's crap. I can confirm it.
Makes me wonder what the problem might be, though.
The causal convolution hasn't been changed, and the input/output shift also hasn't been touched.
Those are the only two places where I would suspect a problem.

ibab:
There was a redundant reshape that seems to have something to do with the problem.
If I remove it, everything seems to work again, but I'm not sure why it makes a difference :/

paperkettle:
Switched to Restructure branch and still seeing immediate drop to 0. 
Training currently seems to work on Master, but generation hits the problems of https://github.com/ibab/tensorflow-wavenet/pull/88/commits/3261ae0095e5bc7e2e3c156e1bb52cd67915ba66 
Running python3 / tensorflow 0.10 / cuda 8.
New in these woods so possible I'm doing something off.!

ibab:
@paperkettle: If the loss is dropping to 0 for you, that's a pretty sure sign that something is still wrong. It's likely that I made a mistake somewhere.

ibab:
I've rebased on `master` and added the changes from #76.

maxhodak:
Hey is this branch getting close to landing? Anything I can do to help / any bugs to track down?

ibab:
Just found the origin of the bug:
I've used `filter_width` as the first dimension for the `1x1` convolutions inside the dilated block instead of `1`.
This meant that they allowed the network to look into future.
The loss now drops at a regular pace.

@maxhodak: It would be great if you could take a look at the code to see whether anything can be expressed in a cleaner way. I'll also go through it.

Next, I'm planning to write a separate `WaveNet` class that manages all the TensorFlow stuff in the background so we can neatly encapsulate a lot of the functionality that's currently in the `train.py` and `generate.py` scripts.
(And this would allow users to use WaveNet more easily for their custom applications)

README.md
```diff
@@ -105,7 +105,10 @@ Passing `--save_every` in addition to `--wav_out_path` will save the in-progress
 python generate.py --wav_out_path=generated.wav --save_every 2000 --samples 16000 model.ckpt-1000
 ```
 
-Fast generation is enabled by default. It uses the implementation from the [Fast Wavenet](https://github.com/tomlepaine/fast-wavenet) repository. You can follow the link for an explanation of how it works. This reduces the time needed to generate samples to a few minutes.
+Fast generation is enabled by default.
+It uses the implementation from the [Fast Wavenet](https://github.com/tomlepaine/fast-wavenet) repository. You can follow the link for an explanation of how it works. This reduces the time needed to generate samples to a few minutes.
+You can follow the link for an explanation of how it works.
```
ibab: This is redundant

test/test_model.py
```diff
@@ -57,15 +56,15 @@ def testEndToEndTraining(self):
         optim = optimizer.minimize(loss, var_list=trainable)
         init = tf.initialize_all_variables()
 
-        max_allowed_loss = 0.1
+        max_allowed_loss = 0.2
```
ibab: This should be changed back

PR #93, opened 2016-09-26
Fix --fast_generation flag in generation script

robinsloan:
The generation script's `fast_generation` flag wasn't working as intended: even when specified on the command line as `False`, the flag was being sent to the model as `True`, due to an unintended behavior of the `argparse` library, [documented here](http://stackoverflow.com/questions/15008758/parsing-boolean-values-with-argparse).

In this patch, I kept the feature's default `true`, as you had it, and added a `disable_fast_generation` flag (which takes no further arguments).

I also updated the README.

Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,14 +105,13 @@ Passing `--save_every` in addition to `--wav_out_path` will save the in-progress
 python generate.py --wav_out_path=generated.wav --save_every 2000 --samples 16000 model.ckpt-1000
 ```
 
-For fast generation:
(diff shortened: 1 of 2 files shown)


ibab:
This is already part of #88, although I have to admit that it's difficult to find something like this in a pull request that's this large:
https://github.com/ibab/tensorflow-wavenet/pull/88/files#diff-bb0b5dadfe137159c585c6ea1e9d5281R22

I've used a slightly different approach of parsing the argument to a boolean.
If you modify your pull request to take the same approach (you can take a look at my PR), then I'll happily merge it, as #88 still needs some work and it would be nice to have this fix on `master`.

ibab:
Awesome, thank you!
I'll merge as soon as travis has finished.

PR #95, opened 2016-09-27
make silence threshold an argument

genekogan:
right now the silence trimming is hardcoded to 0.3 threshold. would be nice to make this adjustable from the args.


ibab:
Looks good, thank you!

PR #96, opened 2016-09-27
Added silence trimming to seed generation

mortont:
Also fixed the case where a negative cut index could occur, which was not intended.


mortont:
I'll squash those commits as well, if you'd prefer.

generate.py
```diff
@@ -85,9 +85,12 @@ def create_seed(filename,
                 quantization_channels,
                 window_size=WINDOW):
     audio, _ = librosa.load(filename, sr=sample_rate, mono=True)
+    audio = audio_reader.trim_silence(audio)
 
     quantized = mu_law_encode(audio, quantization_channels)
-    cut_index = tf.size(quantized) + tf.constant(window_size) - tf.constant(1)
+    cut_index = tf.cond(tf.size(quantized) - tf.constant(window_size) < 0,
+            lambda: tf.constant(window_size),
```
ibab: Shouldn't this be `lambda: tf.size(quantized)`, as we want to use the entire sample.
I think `window_size` would be outside of the sample in this case.

mortont: Good catch, I'll change that.

generate.py
```diff
@@ -85,9 +85,12 @@ def create_seed(filename,
...
+            lambda: tf.size(quantized) - tf.constant(window_size))
```
mortont: This should also be changed to `lambda: tf.constant(window_size)`, come to think of it. That way the seed will be the first `window_size` of audio, or the entire input audio if it's smaller than `window_size`.

ibab: Agreed!

ibab:
Looks good, thank you!

PR #101, opened 2016-09-28
make steps per checkpoint an arg

genekogan:
small addition makes number of steps per checkpoint an argument. is there any reason for it to be so low (50) by default? seems unnecessary to save so often, and it adds a bit overhead to training time.


lemonzi:
Thank you! This was on the to-do's. Other models I've seen save checkpoints, summaries and evals based on time and not on iterations (like, save a checkpoint every hour of training and a summary every 30 seconds), which is easier to manage with highly scalable models, but this is already very helpful.

ibab:
Thanks, this looks good to me!

PR #102, opened 2016-09-28
Minor tweaks in generate.py

lemonzi:
- Only log progress once per second while generating.
- Use the same datestring as train.py
- Save logs in `generate`, not `train`.
- Minor style nits.

### DIFF
--- a/generate.py
+++ b/generate.py
@@ -181,7 +181,8 @@ def main():
         current_sample_timestamp = datetime.now()
         time_since_print = current_sample_timestamp - last_sample_timestamp
         if time_since_print.total_seconds() > 1.:
-            print('Sample {:3<d}/{:3<d}'.format(step + 1, args.samples))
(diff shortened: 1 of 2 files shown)


lemonzi:
Wait, I haven't checked that everything works.

lemonzi:
Ok, I think it works now. Another nice thing to have would be that carriage return trick that overwrite the current line while displaying progress indicators.

lemonzi:
Done.

ibab:
Looks good!
I can merge once travis is happy.

We could use a library like https://github.com/noamraph/tqdm to display nicer progress statistics.
This would be useful for `generate.py`, and maybe even for `train.py`.

lemonzi:
@ibab Nice! I was thinking about writing a generic rate limiter that could limit both number of iterations and time since last call for checkpoint saving, summaries, and logging. If we use tdqm we could still benefit from it for the rest, though.

PR #105, opened 2016-09-28
Do Not Merge - fp16 experiment

jyegerlehner:
I just want to see what the CI server will do when we ask it to run an 16-bit floating point op. I expect it will fail, since there's no (good) way to do fp16 on the CPU.

It runs on my Maxwell GPU though, btw. You can see the loss of significant digits if you run with float16 vs float32.

### DIFF
--- a/test/test_fp16_op.py
+++ b/test/test_fp16_op.py
@@ -1,0 +1,27 @@
+"""Unit tests for fp16 operation"""
+
+import json
+
+import numpy as np
+import tensorflow as tf
+
+from wavenet import WaveNetModel, time_to_batch, batch_to_time, causal_conv
+
+
+class TestFp16(tf.test.TestCase):
+    def testMatMul(self):
+      mat1 = np.identity(5)
+      mat2 = np.identity(5)
+      mat2[1,1] = 0.23456789123456
+      mat2[2,2] = 0.33333333333333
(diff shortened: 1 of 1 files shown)


jyegerlehner:
How about that. It passes. So maybe the Travis CI machines have GPUs? I don't know.

ibab:
@jyegerlehner: `float16` can be emulated on the CPU. For example, this is how `float16` works in numpy. TensorFlow also seems to be able to deal with it. It's not faster, though.

jyegerlehner:
@ibab Right, I know it's not faster in terms of raw flops on a GPU (unless it's Pascal GP100). THere is overhead in that on Maxwell and below the fp16 has to be promoted to fp32 before arithmetic. It's conceivable there could be some speed benefit due to lower memory bandwidth requirements as many ops are bottlenecked by memory latency and bandwidth. 

My motivation for using it would be to reduce the memory footprint of the tensors in half. I expect to do some things that may push the limits of available memory and fp16 would be a bit benefit for that.

ibab:
@jyegerlehner: Sorry, I mean it's not faster when using the CPU.
fp16 is definitely useful if your GPU supports it 👍

Unrelated to this, we could also attempt to evaluate multiple batches on multiple GPUs to speed things up for people who have them.

PR #106, opened 2016-09-28
Add scalar input

ibab:
This adds a `scalar_input` option to the network parameters, as well as an `initial_filter_width` that specifies the size of the filter in the first causal convolution.

Unfortunately, this currently ends up failing the corresponding test case in `test/test_model.py`, because the loss takes way too long to drop when using the scalar input.
Maybe there's a mistake in how I've implemented this, so it might be worth having a closer look.

### DIFF
--- a/wavenet_params.json
+++ b/wavenet_params.json
@@ -1,11 +1,15 @@
 {
     "filter_width": 2,
     "sample_rate": 16000,
-    "dilations": [1, 2, 4, 8, 16, 32, 64, 128, 256,
-                  1, 2, 4, 8, 16, 32, 64, 128, 256],
+    "dilations": [1, 2, 4, 8, 16, 32, 64, 128, 256, 512,
+                  1, 2, 4, 8, 16, 32, 64, 128, 256, 512,
+                  1, 2, 4, 8, 16, 32, 64, 128, 256, 512,
+                  1, 2, 4, 8, 16, 32, 64, 128, 256, 512],
     "residual_channels": 32,
-    "dilation_channels":16,
+    "dilation_channels":32,
     "quantization_channels": 256,
-    "skip_channels": 256,
-    "use_biases": false
+    "skip_channels": 512,
+    "use_biases": false,
+    "scalar_input": true,
+    "initial_filter_width": 32
 }
--- a/wavenet_params.json
+++ b/wavenet_params.json
@@ -1,15 +1,13 @@
 {
     "filter_width": 2,
     "sample_rate": 16000,
-    "dilations": [1, 2, 4, 8, 16, 32, 64, 128, 256, 512,
-                  1, 2, 4, 8, 16, 32, 64, 128, 256, 512,
-                  1, 2, 4, 8, 16, 32, 64, 128, 256, 512,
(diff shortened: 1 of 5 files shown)


jyegerlehner:
Code format?

> test/test_model.py:93:1: E302 expected 2 blank lines, found 1

It will still fail if it's like it was when I last tried scalar input.  I had to bump the test iterations up to 400 and the learn_rate down to 0.005 if I recall.

wavenet_params.json
```diff
@@ -1,11 +1,15 @@
 {
     "filter_width": 2,
     "sample_rate": 16000,
-    "dilations": [1, 2, 4, 8, 16, 32, 64, 128, 256,
-                  1, 2, 4, 8, 16, 32, 64, 128, 256],
+    "dilations": [1, 2, 4, 8, 16, 32, 64, 128, 256, 512,
+                  1, 2, 4, 8, 16, 32, 64, 128, 256, 512,
+                  1, 2, 4, 8, 16, 32, 64, 128, 256, 512,
+                  1, 2, 4, 8, 16, 32, 64, 128, 256, 512],
```
jyegerlehner: Is this related to scalar input? My preference would be for unrelated changes to be in separate PRs.

wavenet_params.json
```diff
@@ -1,11 +1,15 @@
...
     "residual_channels": 32,
-    "dilation_channels":16,
+    "dilation_channels":32,
     "quantization_channels": 256,
-    "skip_channels": 256,
-    "use_biases": false
+    "skip_channels": 512,
+    "use_biases": false,
+    "scalar_input": true,
```
jyegerlehner: Default to "scalar_input": false, until we've proven that it works as well or better?

ibab: Yeah, good point. It's turned on because I was testing it.

ibab:
Since there were no more comments, I'll merge this one now :)

jyegerlehner:
@ibab I've been running an experiment with this, and am close to declaring defeat. I haven't been able to get it to work as well as the quantized input encoding. Loss plateaus quite a bit higher. 

I was going to suggest it wouldn't hurt to merge it, except that it adds code that no-one will use. I have a few things to try before giving up.

By the way generation with scalar inputs didn't work (not supported exception) and haven't looked into it yet at all.

Sorry if I pushed you down this path against your will. The fact that it's not working well bugs me; makes me doubt all my intuitions.

ibab:
@jyegerlehner: No worries, I merged this because I thought it would be useful to have the functionality on `master` for further testing.
Generating audio doesn't work because I haven't yet modified the generator to support scalar input.

Nimeas:
Could someone please explain why `initial_filter_width` defaults to `32` when expecting scalar input instead of `2` which is default for one-hot encoded input?

Thanks

lingerlyn:
I have the same question as above.

PetrochukM:
@ibab Was able to get the loss to drop if the scalar input was centered at zero between the range of [-1, 1]

The authors also mentioned that they did not feed in quantized scalars; they fed in the original scalars before further quantization noise was added.

PR #107, opened 2016-09-29
Add SGD optimizer option

jyegerlehner:
SGD uses less memory than Adam, as it requires fewer copies of activation tensors. So while it doesn't usually work any better than Adam, it could be useful as a fallback if someone runs out of memory training.

### DIFF
--- a/train.py
+++ b/train.py
@@ -33,7 +33,7 @@ SILENCE_THRESHOLD = 0.3
 EPSILON = 0.001
 ADAM_OPTIMIZER = 'adam'
 SGD_OPTIMIZER = 'sgd'
-DEFAULT_SGD_MOMENTUM = 0.9
+SGD_MOMENTUM = 0.9
 
 
 def get_arguments():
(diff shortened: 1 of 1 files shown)


train.py
```diff
@@ -77,6 +80,13 @@ def get_arguments():
                         default=SILENCE_THRESHOLD,
                         help='Volume threshold below which to trim the start '
                         'and the end from the training set samples.')
+    parser.add_argument('--optimizer', type=str, default='adam',
```
jyegerlehner: 'adam' should ADAM_OPTIMIZER

train.py
```diff
@@ -30,6 +30,9 @@
 SAMPLE_SIZE = 100000
 L2_REGULARIZATION_STRENGTH = 0
 SILENCE_THRESHOLD = 0.3
+ADAM_OPTIMIZER = 'adam'
+SGD_OPTIMIZER = 'sgd'
+DEFAULT_SGD_MOMENTUM = 0.9
```
ibab: We can call it `SGD_MOMENTUM`, as none of the other defaults have a `DEFAULT` prefix

nakosung:
@jyegerlehner  Do you use SGD for 500ms receptive field?

jyegerlehner:
nakosung: No I haven't been yet.

> (I found your generated output on sound cloud and it sounds promising!)

[This one](https://soundcloud.com/user-731806733/p280-500-msec-receptive-field-b)? That was after some training with 0.00002 learning rate , and the silence trimming threshold removed. The problem at first seemed we were generating an un-ending stream of vowels and consonants, so I reduced the silence trimming threshold. Then I started getting too much silence, such as the long quiet voids in that clip. I generated a minute long sample and it was mostly silence with a few single-syllables of speech here and there. But when there is speech, it sounds pretty good.

ibab:
@jyegerlehner: We should be able to control that by framing the samples with a fixed amount of silence.

ibab:
LGTM!

PR #108, opened 2016-09-29
Import print() from future for 2.x

maxhodak:
Calling a keyword argument on `print` is a syntax error in Python 2.x.

### DIFF
--- a/generate.py
+++ b/generate.py
@@ -1,4 +1,5 @@
 from __future__ import division
+from __future__ import print_function
 
 import argparse
 from datetime import datetime
@@ -185,7 +186,7 @@ def main():
         time_since_print = current_sample_timestamp - last_sample_timestamp
         if time_since_print.total_seconds() > 1.:
             print('Sample {:3<d}/{:3<d}'.format(step + 1, args.samples),
-                  end='\r')
+               end='\r')
             last_sample_timestamp = current_sample_timestamp
 
         # If we have partial writing, save the result so far.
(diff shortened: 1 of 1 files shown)


jyegerlehner:
Interesting. I've been running generate.py in python 2.7 and it's been working. How can that be? I don't know much about python.

maxhodak:
I don't know. This recently broke with one of the diffs landed today. Assuming you were on the same code, are you sure you're using 2.7? Virtualenv can sometimes cause you to use a different version of python if you're using that.

mecab:
@jyegerlehner @maxhodak 

https://github.com/ibab/tensorflow-wavenet/blob/4dafacbc42024797bf022eca6ca4f87d8c898e52/generate.py#L189 had been broke in Python2.

In Python2, `print` was a statement and not a function (so you could write `print "foo"`). In Python3, `print` is completely function, and can take some optional parameters, such as `end`.

`from __future__ import print_function` magic injects Python3 style `print` function into Python2. So it enables Python2 to run the newer codes.

lemonzi:
Why didn't Travis complain? Aren't the generate tests run in CI as well?

mecab:
@lemonzi 
No, it looks current CI only uses `test/test_generation.py` and check the model's stability, not `generate.py`'s. If we want to guarantee the stability of `generate.py` itself, we might need more tests.

maxhodak:
I think just importing something from generate.py in anything used by the tests should be sufficient to catch syntax errors

mecab:
Right. Introducting static lint such as pyflakes is another option.

lemonzi:
+1 for a linter

PR #113, opened 2016-09-29
Scalar Input: switch to MomentumOptimizer. Get tests passing.

jyegerlehner:
This PR is against ibab's add-scalar-input branch.

I noticed when I tried to let the test run more iterations, the loss would be dropping nicely and then suddenly pop up to a large number. It would do this for all three models, one-hot, with-biases, scalar-input. I tried adding regularization but that didn't help. Then I replaced AdamOptimizer with MomentumOptimizer and the problem vanished. Seems to be an instability with Adam?

Anyway, these tests were passing on my machine now.

### DIFF
--- a/test/test_model.py
+++ b/test/test_model.py
@@ -62,7 +62,7 @@ class TestNet(tf.test.TestCase):
         init = tf.initialize_all_variables()
 
         max_allowed_loss = 0.1
-        val = max_allowed_loss
+        loss_val = max_allowed_loss
         initial_loss = None
         with self.test_session() as sess:
             sess.run(init)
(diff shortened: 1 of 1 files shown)


test/test_model.py
```diff
@@ -51,20 +55,22 @@ def testEndToEndTraining(self):
...
-        optimizer = tf.train.AdamOptimizer(learning_rate=0.02)
+        optimizer = tf.train.MomentumOptimizer(learning_rate=LEARN_RATE,
+                                               momentum=MOMENTUM)
         trainable = tf.trainable_variables()
         optim = optimizer.minimize(loss, var_list=trainable)
         init = tf.initialize_all_variables()
 
         max_allowed_loss = 0.1
-        loss_val = max_allowed_loss
+        val = max_allowed_loss
```
jyegerlehner: Oops. Don't know what happened here.

ibab:
Ah, I never even considered that it could be caused by the optimizer!

PR #115, opened 2016-09-30
Default the silence trimming threshold in exactly 1 place

jyegerlehner:
I thought I was changing the silence trimming threshold, but it appears in more than one place in the code. I was changing the wrong one. Only the default specified by the command line args has real effect.

So this PR removes the bogus defaults for the silence trim threshold that appear in the code, leaving just the one specified in command-line args.

Edit: I've also added a commit to allow us to skip silence trimming completely.


lemonzi:
This change breaks a bit the re-usability of audio_reader, but since the default threshold was arbitrary anyways I think it's fair to do something about it. It would make more sense to not trim anything by default though, rather than throwing an error. So, if the `threshold` is `None`, just skip the trimming.

lemonzi:
LGTM, thanks!

Issue #98, opened 2016-09-28
Calculation of the loss


JesseYang:
In each step, the calculation of the loss is based on the cross entropy of the input value and the predicted value. Each data point in the predicted value is predicted based on its receptive field in the input value. During prediction, the input value is padded on the left with many 0s to implement the casual convolution. Thus, the beginning predicted data points are actually predicted based on those padding 0s, and may not be the same as the input values even when the model is trained for many steps. Is this the reason that why the loss drops to around 2 and cannot drop lower? Maybe it is more reasonable to calculate the loss not from the beginning data point, but from the N-th point where N is the receptive field.

nakosung:
If so, final output layer should be truncated not to report false losses? I think it might do harm with network parameters.

nakosung:
Implementation in other repo: https://github.com/basveeling/wavenet/blob/master/wavenet.py#L491

jyegerlehner:
@JesseYang 
OK I think I finally understand what you are saying. I think you are right. I'll try to put it in my own words: when computing the discrete prob distribution for the next sample, the net needs (and usually has) all the prior values available to it in order to compute that next predicted sample. The exception is: any time in the time history less than N samples from the start (where N is the receptive field width), we are penalizing the net for failing to predict the output, when we haven't given it all the inputs it needs. In fact, the first sample it predicts has no input values at all available to it.  Yeah. 

So the concern is not that the displayed loss number is a bit inflated. Rather, that we back-propagate errors and adjust net weights and biases for cases where we haven't given the net all the inputs it needs. That seems it would have a distorting, corrupting effect.

Most clips are several seconds long, and the net may have a, say, 300 msec receptive field. So it's a relatively small portion of the weight updates that have this corruption, if it can be called that. 

@nakosung Is your code there a proposed modification of model.py loss method, whereby we exclude the first N samples from the loss calculation? If so I think I agree that would be a fix. Though I'm not familiar with tf.slice and can't judge its correctness yet.

nakosung:
@jyegerlehner Most fast way would be using masking out first N samples from calculated loss just before reducung it.

akademi4eg:
Seems like this is resolved now with merge of PR 191.

Fjolnir-Dvorak:
It actually appeared yesterday again in a training session on the newest commit. I will try to pin it down. The only problem: I trained without metadata and log. I will add a proof of failure this week.

PR #118, opened 2016-10-03
[WIP] Compute loss for outputs only where receptive field is filled

jyegerlehner:
This is my attempt at a fix for [issue 98](https://github.com/ibab/tensorflow-wavenet/issues/98), using nakosung's suggested solution.

I've merged it into a branch I'm training on, and there is not an immediate drop to lower reported loss. Though it might be learning a bit faster. Hard to say. At least it doesn't appear to have broken anything.

### DIFF
--- a/test/test_loss.py
+++ b/test/test_loss.py
@@ -3,31 +3,31 @@ from __future__ import print_function
 import json
 import numpy as np
 import tensorflow as tf
-from wavenet import MakeLoss
+from wavenet import MakeLoss, ComputeReceptiveFieldSize
 
 
 def MakeLogits():
-    # 3 channels (quantization levels)
-    # 5 time steps in duration
     # 1 batch item
-    array = np.array([ [ [  5.0, -10.0, -10.0],
-                         [ 10.0, -10.0, -10.0],
-                         [  0.0,  10.0,   0.0],
-                         [-10.0,  10.0, -10.0],
-                         [-10.0,  10.0, -20.0] ] ])
+    # 5 time steps in duration
+    # 3 channels (quantization levels)
+    array = np.array([[[5.0,   -10.0, -10.0],
+                       [10.0,  -10.0, -10.0],
+                       [0.0,    10.0,   0.0],
+                       [-10.0,  10.0, -10.0],
+                       [-10.0,  10.0, -20.0]]])
 
     return tf.convert_to_tensor(array, tf.float32)
 
 
 def MakeLabels():
-    # 3 channels (quantization levels)
-    # 5 time steps in duration
     # 1 batch item
-    array = np.array([ [ [ 0.0, 1.0, 0.0],
-                         [ 0.0, 0.0, 1.0],
-                         [ 0.0, 0.0, 1.0],
-                         [ 0.0, 0.0, 1.0],
-                         [ 0.0, 0.0, 1.0] ] ])
+    # 5 time steps in duration
+    # 3 channels (quantization levels)
+    array = np.array([[[0.0, 1.0, 0.0],
+                       [0.0, 0.0, 1.0],
+                       [0.0, 0.0, 1.0],
+                       [0.0, 0.0, 1.0],
+                       [0.0, 0.0, 1.0]]])
     return tf.convert_to_tensor(array, tf.float32)
 
 
@@ -35,38 +35,48 @@ class TestLoss(tf.test.TestCase):
     def setUp(self):
         pass
 
-
     def testZeroReceptiveFieldSize(self):
         logits = MakeLogits()
         labels = MakeLabels()
 
-        loss = MakeLoss(labels, logits, quantization_channels = 3,
-                                     receptive_field_size = 0)
+        loss = MakeLoss(labels, logits, quantization_channels=3,
+                        receptive_field_size=0)
         with self.test_session() as sess:
             loss_array = sess.run([loss])
-        expected =np.array([[ 15.00000095, 20.0, 10.0, 20.0, 30.0]])
+        expected = np.array([[15.0, 20.0, 10.0, 20.0, 30.0]])
         EPSILON = 0.001
         np.testing.assert_allclose(expected, loss_array, rtol=EPSILON)
-
 
     def testNonZeroReceptiveFieldSize(self):
         logits = MakeLogits()
         labels = MakeLabels()
 
-        loss = MakeLoss(labels, logits, quantization_channels = 3,
-                                     receptive_field_size = 2)
+        loss = MakeLoss(labels, logits, quantization_channels=3,
+                        receptive_field_size=2)
         with self.test_session() as sess:
             loss_array = sess.run([loss])
 
         # Since we truncate the first receptive_field time elements from the
         # the tensors used to compute the loss, the losses should exclude the
-        # first two (since receptive_field_size is set to 2).
-        expected =np.array([[ 10.0, 20.0, 30.0]])
+        # first two (since receptive_field_size is set to 2) that we would
+        # otherwise have.
+        expected = np.array([[10.0, 20.0, 30.0]])
 
         EPSILON = 0.001
         np.testing.assert_allclose(expected, loss_array, rtol=EPSILON)
 
+    def testComputeReceptiveFieldSize_4(self):
+        dilations = [1, 2, 4]
+        computed_size = ComputeReceptiveFieldSize(dilations)
+        self.assertEqual(computed_size, 8)
+
+    def testComputeReceptiveFieldSize_256_256_256(self):
+        dilations = [1, 2, 4, 8, 16, 32, 64, 128, 256,
+                     1, 2, 4, 8, 16, 32, 64, 128, 256,
+                     1, 2, 4, 8, 16, 32, 64, 128, 256]
+        computed_size = ComputeReceptiveFieldSize(dilations)
+        self.assertEqual(computed_size, 512 + 511 + 511)
+
 
 if __name__ == '__main__':
     tf.test.main()
-
--- a/test/test_loss.py
+++ b/test/test_loss.py
@@ -1,0 +1,72 @@
+"""Unit tests for the loss function."""
+from __future__ import print_function
+import json
+import numpy as np
+import tensorflow as tf
+from wavenet import MakeLoss
+
+
+def MakeLogits():
+    # 3 channels (quantization levels)
+    # 5 time steps in duration
+    # 1 batch item
+    array = np.array([ [ [  5.0, -10.0, -10.0],
+                         [ 10.0, -10.0, -10.0],
+                         [  0.0,  10.0,   0.0],
+                         [-10.0,  10.0, -10.0],
+                         [-10.0,  10.0, -20.0] ] ])
+
+    return tf.convert_to_tensor(array, tf.float32)
+
+
+def MakeLabels():
+    # 3 channels (quantization levels)
+    # 5 time steps in duration
+    # 1 batch item
+    array = np.array([ [ [ 0.0, 1.0, 0.0],
+                         [ 0.0, 0.0, 1.0],
+                         [ 0.0, 0.0, 1.0],
+                         [ 0.0, 0.0, 1.0],
+                         [ 0.0, 0.0, 1.0] ] ])
+    return tf.convert_to_tensor(array, tf.float32)
+
+
+class TestLoss(tf.test.TestCase):
+    def setUp(self):
+        pass
+
+
+    def testZeroReceptiveFieldSize(self):
+        logits = MakeLogits()
+        labels = MakeLabels()
+
+        loss = MakeLoss(labels, logits, quantization_channels = 3,
+                                     receptive_field_size = 0)
+        with self.test_session() as sess:
+            loss_array = sess.run([loss])
+        expected =np.array([[ 15.00000095, 20.0, 10.0, 20.0, 30.0]])
+        EPSILON = 0.001
+        np.testing.assert_allclose(expected, loss_array, rtol=EPSILON)
+
+
+    def testNonZeroReceptiveFieldSize(self):
+        logits = MakeLogits()
+        labels = MakeLabels()
+
+        loss = MakeLoss(labels, logits, quantization_channels = 3,
+                                     receptive_field_size = 2)
+        with self.test_session() as sess:
+            loss_array = sess.run([loss])
+
+        # Since we truncate the first receptive_field time elements from the
+        # the tensors used to compute the loss, the losses should exclude the
+        # first two (since receptive_field_size is set to 2).
+        expected =np.array([[ 10.0, 20.0, 30.0]])
+
+        EPSILON = 0.001
+        np.testing.assert_allclose(expected, loss_array, rtol=EPSILON)
+
+
+if __name__ == '__main__':
+    tf.test.main()
+
--- a/test/test_loss.py
+++ b/test/test_loss.py
@@ -1,45 +1,38 @@
-"""Unit tests for the loss function."""
+"""Unit tests for the make_loss function."""
+
 from __future__ import print_function
 import json
 import numpy as np
+
 import tensorflow as tf
-from wavenet import make_loss, compute_receptive_field_size
+
+from wavenet import make_loss, compute_receptive_field
 
 
-def make_logits():
-    # 1 batch item
-    # 5 time steps in duration
-    # 3 channels (quantization levels)
-    array = np.array([[[5.0,   -10.0, -10.0],
-                       [10.0,  -10.0, -10.0],
-                       [0.0,    10.0,   0.0],
-                       [-10.0,  10.0, -10.0],
-                       [-10.0,  10.0, -20.0]]])
-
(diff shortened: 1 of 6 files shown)


ibab:
I wonder how this affects generation, as the network now never learns the case where its receptive field is cut short by the left boundary.
Could you check whether we can still generate sensible audio with this change?

nakosung:
This fix eliminates unnecessary dead zone by zeroing.

Causal convolution processes whole layer, so if the layer contains 'dead zone' its gradient should remain as zero not to distort convolution parameters. 

IMO, this removes unused left part of model and generator reads the rightmost sample, so I think there should be no difference in output.

jyegerlehner:
@ibab Good point; will check generation.

jyegerlehner:
@ibab I checked by training with this branch for a day or so, and it generates the same kind of stuff as before, as far as I can (subjectively) tell.

nakosung:
image

I think it might be faster to use masking instead of slice.

jyegerlehner:
Are you referring to tf.boolean_mask or tf.sequence_mask? I would guess those are slower.

nakosung:
Element-wise multiplication by selective zero/one. :) (0{receptifive field size} 1{N - receptive field size})

jyegerlehner:
So we'd first have to create a tensor of the same size as the loss tensor, fill it with ones and zeros, elementwise multiply and store the result back to global memory? Pretty sure that's slower than just copying out the part we want with slice.

nakosung:
@jyegerlehner ones and zeros are initialized once as a global constant vector. element wise multiplication can be done in-place(I don't know how tensorflow handles this *mul* internally with CUDA). 

btw, after seeing the diagram above, I think the most fast and clean way is to shift each layers. (-1, -3, -7, -15).

jyegerlehner:
> ones and zeros are initialized once as a global constant vector. element wise multiplication can be done in-place(I don't know how tensorflow handles this mul internally with CUDA). 

Right, so that looks to me like it uses at least as much memory, needs to retrieve both the mask and starting tensor from memory, perform a floating point op on each element, and store the result to global memory. I don't see how it could be faster.

> after seeing the diagram above, I think the most fast and clean way is to shift each layers. (-1, -3, -7, -15).

Yikes, well you may be seeing something I don't. I can't imagine a way to do that that is anywhere near as simple as what he have now, nor do I see any efficiency gain. Seems like you'd have to visit every tensor in the hierarchy to pull that off. I could easily be wrong. If someone wants to produce some code to suggest as simpler or more efficient than what we have now, that would be good. Please type real slow though so I can follow :). I don't see how to do it. Simplicity (and therefore understandability) of the code is a prime consideration. I don't think what we're doing now has any noticeable impact on speed, and is only a couple lines of code.

nakosung:
@jyegerlehner Nevermind. :) Speed gain would be minimal comparing to current frame size. It would be a kind of early micro optimization. I'm satisfied with fix for possibly-incorrect back propagation from dead zone.

lemonzi:
General comment: it looks you're using camel case for auxiliary functions as Google does, and so far we've been using snake case for functions and camel case for classes. It would be good to use snake case for the time being to be consistent, we can always change it later. There are also minor style nits, such as docstrings with an inconsistent format or missing newlines.

We should also try to keep `model.py` as lean and clear as possible, as it's already quite big. I don't know if it would be better to move some parts to `ops.py`, or turn `MakeLoss` into a method that reads from `self.quantization_channels` and order the methods in some sensible way.

jyegerlehner:
@lemonzi OK thanks; will fix the case thing.

> docstrings with an inconsistent format or missing newlines.

Any chance you could mark those with review comments? I'm bad with style and noob at python and docstrings.

jyegerlehner:
If we give the net an audio clip that's shorter than the receptive field that causes the tf.slice operation to fail. So we need to detect that at runtime and skip the audio if, after silence trimming, it is shorter than the receptive field. 

A similar failure is when sample_size is specified. We need to ensure that sample_size is greater than the receptive field.

I put logic in to do those things. Anyone care to review and check those changes?

ibab:
I'm still not sure whether it's a good idea to aggressively cut away the region that doesn't have a fully covered receptive field. After all, the network is still able to make use of all leaves of the evaluation tree that land inside the waveform (see figure 3 from the paper for what I mean by evaluation tree).

This also adds quite a bit of complexity, as we need to compute the size of the receptive field and handle inputs that are too small correctly (Whereas before, the network might have been perfectly capable of dealing with small inputs).

I'd be easily persuaded if we can find something that's improved by this, though (like faster training).

nakosung:
@ibab I think cutting dead region should be done to ensure strictness of 'the loss'. To make unused predictions generated by network not to contribute to gradients, they should be silenced by any means. 

Current `SAMPLE_SIZE` is 100000(~4.5s), which is just 9 times larger than 500ms (~receptivie field size). So if we don't cut out dead zone, I think we are using `true-loss * (1 + 0.1 * noise)`. (In fact, it wouldn't be so  bad for training to add some noise. But it might make convergence harder, IMO.)

wavenet/model.py
```diff
@@ -18,6 +18,64 @@ def create_bias_variable(name, shape):
...
+
+
+def compute_receptive_field_size(dilations):
+    '''Given the list of dilations, return the receptive field size of the
+    net.
+
+    Arguments:
+        dilations: List of dilation values used by the dilated convolutions.
+
+    WARNING: this implemenation is only exactly correct for dilations list
```
lemonzi: This should go before the argument list.

wavenet/model.py
```diff
@@ -18,6 +18,64 @@ def create_bias_variable(name, shape):
...
+    loss = tf.nn.softmax_cross_entropy_with_logits(
+        prediction, tf.reshape(labels, [-1, quantization_channels]))
+    return loss
+
+
+def compute_receptive_field_size(dilations):
+    '''Given the list of dilations, return the receptive field size of the
+    net.
+
+    Arguments:
```
lemonzi: The Google style guide suggests "Args:" instead.

wavenet/model.py
```diff
@@ -18,6 +18,64 @@ def create_bias_variable(name, shape):
     return tf.Variable(initializer(shape=shape), name)
 
 
+def make_loss(labels, logits, quantization_channels, receptive_field_size):
+    '''Create the loss function from the net's raw output (softmax logits)
+    and the labels (as probs).
+
+    Arguments:
```
lemonzi: The Google style guide suggests "Args:" instead.

wavenet/model.py
```diff
@@ -18,6 +18,64 @@ def create_bias_variable(name, shape):
...
+        labels: Target probability tensor, shape: [batch size,
+                duration in time steps, channels (quantization levels)]
+
```
lemonzi: No need for these empty lines.

wavenet/model.py
```diff
@@ -18,6 +18,64 @@ def create_bias_variable(name, shape):
     return tf.Variable(initializer(shape=shape), name)
 
 
+def make_loss(labels, logits, quantization_channels, receptive_field_size):
+    '''Create the loss function from the net's raw output (softmax logits)
+    and the labels (as probs).
+
+    Arguments:
+        labels: Target probability tensor, shape: [batch size,
+                duration in time steps, channels (quantization levels)]
```
lemonzi: Missing period at the end.

lemonzi: Try to get a formatting that breaks at a better point, like

```
Target probability tensor, with shape:
[batch_size, duration, quantization_levels],
where duration is in time steps.
```

train.py
```diff
@@ -170,6 +170,18 @@ def validate_directories(args):
     }
 
 
+def check_sample_size(dilations, sample_size):
+    '''Check that the sample size is greater than the receptive field size and
```
lemonzi: It should be double quotes, but this is not your fault -- I'll send a refactor PR when this is merged changing all of them.

lemonzi: The first sentence should also fit in a line, but this is also not enforced elsewhere. For instance, you don't need to talk about the assertion yet. Or, even better, name it `assert_sample_size` so that what it does is obvious.

You should then add a section with the Args, and a Raises section where you specify which exception is raised.

train.py
```diff
@@ -170,6 +170,18 @@ def validate_directories(args):
...
+       raise an exception if not.'''
+
+    receptive_field_size = compute_receptive_field_size(dilations)
+    if sample_size is not None:
+        if sample_size <= receptive_field_size:
```
lemonzi: Why not `and`?

train.py
```diff
@@ -170,6 +170,18 @@ def validate_directories(args):
...
+            raise ValueError('Sample size {} must be greater than the '
+                             'receptive field size {}.'. \
```
lemonzi: These kind of line breaks are very ugly, I'm surprised pep8 hasn't complained about it. I would put `format(` here and the arguments in the next line if they don't fit, or something like that.

train.py
```diff
@@ -200,12 +212,17 @@ def main():
         # zero.
         silence_threshold = args.silence_threshold if args.silence_threshold > \
                                                       EPSILON else None
+        check_sample_size(wavenet_params['dilations'], args.sample_size)
+        receptive_field_size = \
```
lemonzi:
```
receptive_field_size = compute_receptive_field_size(
    wavenet_params['dilations'])
```
It feels a bit too verbose, though.

wavenet/__init__.py
```diff
@@ -1,4 +1,4 @@
-from .model import WaveNetModel
+from .model import (WaveNetModel, make_loss, compute_receptive_field_size)
```
lemonzi: Aren't the parentheses only needed when there is a line break?

train.py
lemonzi: Why single quotes? Is it to be consistent with the rest of the file? If that's the case, let me know and I'll fix it in the PR as well.

wavenet/audio_reader.py
```diff
@@ -94,18 +96,28 @@ def thread_main(self, sess):
...
-                    # Cut samples into fixed size pieces
-                    buffer_ = np.append(buffer_, audio)
-                    while len(buffer_) > self.sample_size:
-                        piece = np.reshape(buffer_[:self.sample_size], [-1, 1])
+                    if audio.size <= self.receptive_field_size:
+                        print("Warning: {} was ignored since, after silence "
+                              "trimming, it was shorter than the receptive "
+                              "field of the net. ".format(filename))
+
+                # If the trimmed audio is less than the receptive field size,
```
lemonzi: The audio is not "less", its length is. You could use "shorter" instead.

lemonzi: The "So" is unnecessary.

wavenet/audio_reader.py
```diff
@@ -94,18 +96,28 @@ def thread_main(self, sess):
...
+                # then tf.slice performed by loss will fail. So skip the file
+                # if this is the case.
+                if audio.size > self.receptive_field_size:
+                    if self.sample_size:
```
lemonzi: I know this is not your code, but since we're at it -- sample_size is not a boolean, so we should explicitly test for `is not None`, or `> 0`, or whatever appropriate.

wavenet/model.py
```diff
@@ -18,6 +18,64 @@ def create_bias_variable(name, shape):
...
+         1, 2, 4, 8, ... 2**m,
+         ...
+         1, 2, 4, 8, ... 2**m]
+    That is, it assumes the most usual form, where the dilations list ends
+    with the largest dilation value, filter width is 2 etc. A formula
+    that is generally correct is deferred for now.
+
+    Returns the receptive field size in sample counts.
+    '''
+    M = max(dilations)
```
lemonzi: max_dilation

wavenet/model.py
```diff
@@ -18,6 +18,64 @@ def create_bias_variable(name, shape):
...
+    N = dilations.count(M)
```
lemonzi: num_stacks, which reflects your assumption

test/test_loss.py
```diff
@@ -0,0 +1,82 @@
+"""Unit tests for the loss function."""
```
lemonzi: Missing newline after the header.

lemonzi: Do you mean `make_loss`?

test/test_loss.py
```diff
@@ -0,0 +1,82 @@
...
+from __future__ import print_function
+import json
+import numpy as np
```
lemonzi: Missing newline before external dependencies

test/test_loss.py
```diff
@@ -0,0 +1,82 @@
...
+import tensorflow as tf
+from wavenet import make_loss, compute_receptive_field_size
```
lemonzi: Missing newline before internal dependencies

test/test_loss.py
```diff
@@ -0,0 +1,82 @@
...
+    # 3 channels (quantization levels)
+    array = np.array([[[0.0, 1.0, 0.0],
+                       [0.0, 0.0, 1.0],
+                       [0.0, 0.0, 1.0],
+                       [0.0, 0.0, 1.0],
+                       [0.0, 0.0, 1.0]]])
+    return tf.convert_to_tensor(array, tf.float32)
+
+
+class TestLoss(tf.test.TestCase):
```
lemonzi: Define a class for each function that you test (make_loss and compute_receptive_field_size)

test/test_loss.py
```diff
@@ -0,0 +1,82 @@
...
+        # Since we truncate the first receptive_field time elements from the
+        # the tensors used to compute the loss, the losses should exclude the
+        # first two (since receptive_field_size is set to 2) that we would
+        # otherwise have.
+        expected = np.array([[10.0, 20.0, 30.0]])
+
+        EPSILON = 0.001
+        np.testing.assert_allclose(expected, loss_array, rtol=EPSILON)
+
+    def test_compute_receptive_field_size_4(self):
```
lemonzi: Use a name that reflects the intention, not the implementation. `compute_receptive_field_size` should be give by the class name. Instead of `4`, use `_single_stack`

test/test_loss.py
```diff
@@ -0,0 +1,82 @@
...
+        dilations = [1, 2, 4]
+        computed_size = compute_receptive_field_size(dilations)
+        self.assertEqual(computed_size, 8)
+
+    def test_compute_receptive_field_size_256_256_256(self):
```
lemonzi: Same here: use `multiple_stacks`.

test/test_loss.py
```diff
@@ -0,0 +1,82 @@
+"""Unit tests for the loss function."""
+from __future__ import print_function
+import json
+import numpy as np
+import tensorflow as tf
+from wavenet import make_loss, compute_receptive_field_size
+
+
+def make_logits():
```
lemonzi: Instead of defining these functions, you could just create the tensors in the `setUp` method and save them as class attributes.

test/test_loss.py
```diff
@@ -0,0 +1,82 @@
...
+    array = np.array([[[0.0, 1.0, 0.0],
+                       [0.0, 0.0, 1.0],
+                       [0.0, 0.0, 1.0],
+                       [0.0, 0.0, 1.0],
+                       [0.0, 0.0, 1.0]]])
+    return tf.convert_to_tensor(array, tf.float32)
+
+
+class TestLoss(tf.test.TestCase):
+    def setUp(self):
```
ibab: You can leave out the `setUp`method if you don't need to perform any setup.

jyegerlehner:
@ibab, @nakosung 
I'm pondering the two sides of this. ibab surprises us and argues it's a feature not a bug. My view of it now pops back and forth like a Rorschach test. 

I generated some audio after training for a while with the skipped receptive field. I thought it sounded better, more like human speech in the first few seconds. But I'm not sure I trust my own subjectivity, especially when I want something to work.

Loss was still higher than what I had achieved with a previous model so I resumed training. Then generated again and got an all-white-noise output. It's a tip revision where I merge various branches into it that I want to experiment with. Somewhere along the way I must have pulled in something that broke and am trying to figure that out. I've gotten myself confused bouncing around between so many development branches on a couple of machines. Will come back if I can get it sorted out and produce a sample to listen to.

@lemonzi Thanks for all that. Will get to it eventually.

jyegerlehner:
Here are a couple generated clips after training with this PR.

https://soundcloud.com/user-731806733/tensorflow-wavenet-generated-42
https://soundcloud.com/user-731806733/wavenet-tensorflow-generated-43

I think I'll resume training with silence trimming turned off; I think I've observed that it tends to generate run-on words without pauses like these when silence trimming is always on. Or it could be due to shorter receptive field (300 msec) as compared to 500 msec of earlier clips that had the pauses.

By the way these were generated with "slow" generation as fast generation just generates static.

Zeta36:
@jyegerlehner 

Look at this #126 for the fast generation problem.

jyegerlehner:
My experiments didn't show any obvious improvement from this. And it makes the code substantially more complicated. And might be the opposite of what we want if we were to implement a longer-range context stack. We can always re-open and revisit if someone thinks we need this.

Issue #77, opened 2016-09-23
Change learning rate during training


r-zemblys:
Would it be helpful to add ability to manually change learning rate during training? I have naive implementation of this feature [here](https://github.com/r-zemblys/tensorflow-wavenet/tree/dev).
![lr2](https://cloud.githubusercontent.com/assets/8914323/18787185/694e945e-81aa-11e6-9a0f-b47b12486e4b.png)
![lr1](https://cloud.githubusercontent.com/assets/8914323/18787184/694bd110-81aa-11e6-97c9-3665a5351b0f.png)

lemonzi:
The standard approach is to add 2 parameters: a number of iterations during which the learning rate is kept constant, and a multiplier that is applied after that for exponential decay. This forces the network to adjust to the closest local minima when training is running out of iterations rather than keep exploring the space.

The best way of changing it manually is stopping the training and restoring it from a checkpoint with a different learning rate.

I think TensorFlow has a built-in implementation of this exponential decay, or at least plenty of examples in the published models; it would be indeed very helpful to have it.

jyegerlehner:
@r-zemblys  
Yes please. Echoing lemonzi, I'd say your particular schedule of learning rate (in the plot) looks a little odd :). I'm guessing you are showing the "square wave" learning rate schedule to show it can be adjusted at will, not that you recommend that particular schedule?

@lemonzi 
> The standard approach is to add 2 parameters: a number of iterations during which the learning rate is kept constant, and a multiplier that is applied after that for exponential decay.

Hmm.. I'm not aware of that standard. It may be a good strategy. I've seen people just using a single step-change from an initial high learning rate to an order-of-magnitude lower learning rate later. I'm not saying that's better. But I wouldn't want to hard-wire to just the "constant-followed-by-exponential-decay" approach.

>  The best way of changing it manually is stopping the training and restoring it from a checkpoint with a different learning rate.

Do you say that because you think there's something wrong with the technique in r-zembly's dev branch? I've done it the way you describe, but I like his approach of having a learning_rate placeholder and feeding it a value you can set any way you want as training progresses.

Perhaps a configurable approach would be to have a learning_rate_schedule in wavenet_params.json with a list of (iterations, learn_rate) tuples, and use that to compute learn_rate fed to the placeholder.

lemonzi:
[...]

PR #120, opened 2016-10-05
Rmsprop optimizer

jyegerlehner:
DeepMind is using RMS prop in [Video Pixel Networks](https://arxiv.org/pdf/1610.00527v1.pdf)  as nakosung pointed out over [there](https://github.com/ibab/tensorflow-wavenet/issues/77). Video Pixel Networks are related to Wavenet.

I figured I would try RMS prop out and see if a miracle ensues.

### DIFF
--- a/generate.py
+++ b/generate.py
@@ -113,7 +113,9 @@ def main():
         dilation_channels=wavenet_params['dilation_channels'],
         quantization_channels=wavenet_params['quantization_channels'],
         skip_channels=wavenet_params['skip_channels'],
-        use_biases=wavenet_params['use_biases'])
+        use_biases=wavenet_params['use_biases'],
+        scalar_input=wavenet_params['scalar_input'],
+        initial_filter_width=wavenet_params['initial_filter_width'])
 
     samples = tf.placeholder(tf.int32)
 
--- a/test/test_model.py
+++ b/test/test_model.py
@@ -7,34 +7,38 @@ import tensorflow as tf
 
 from wavenet import WaveNetModel, time_to_batch, batch_to_time, causal_conv
 
+SAMPLE_RATE_HZ = 2000.0  # Hz
+TRAIN_ITERATIONS = 400
+LEARN_RATE = 0.02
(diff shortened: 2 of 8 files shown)


lemonzi:
Does it help? 

It would be nice to have a more generic optimizer selector rather than all these hardcoded constants. How about a dictionary that maps names to functions, and the argument parser is given its `.keys()`? I don't know what to do with the momentum though. Maybe we need a separate function or class that handles this. Or, at least, WaveNet could export a list with the supported optimizers rather than read the individual globals from the outside.

lemonzi:
This is on top of #107, right?

jyegerlehner:
Yes, on top of 107.

>  more generic optimizer selector rather than all these hardcoded constants. How about a dictionary

Yes, that's a better idea.

jyegerlehner:
> Does it help? 

I haven't started running it yet. I thought I could just switch the optimizer and resume training. But it fails to load the checkpoint because there are tensors specific to the optimizer it tries to load and can't find. I think we must be saving more than just the trainable vars in our checkpoints. Is that intended, or can we change that?

jyegerlehner:
Here's comparison training on speaker p280. Orange line is SGD (learn rate = 0.01, momentum = 0.9), aqua is adam (learn rate = 0.001), purple is RMS Prop (learn rate = 0.001, momentum=0.9).

screenshot from 2016-10-07 00 25 47

lemonzi:
There's something wrong with the diff, it looks like your branch undoes PR's that have already been applied and then it re-applies them. If will not have any effect because of the squash, but it makes it hard to see the changes. Could you try to fix it? I believe this is GitHub's fault, for being too smart with the squash merges. You may have to undo your internal merge, rebase the old branch to the current master, and merge again.

ccienfall:
@jyegerlehner, have Deep mind or others open the Video Pixel Networks' source codes? Recently I am reading this paper. Thanks.

jyegerlehner:
@ccienfall Not as far as I know.

PR #121, opened 2016-10-05
Save and load only trainable vars

jyegerlehner:
Currently we save, and try to load, all variables including ones particular to the optimizer. This prevents one from changing the optimizer, and resuming training from a saved checkpoint.

I found that if we only save the trainable vars then I can change the optimizer and resume training from a saved checkpoint.

Is there some reason why saving all the variables is preferable?

### DIFF
--- a/train.py
+++ b/train.py
@@ -227,7 +227,7 @@ def main():
     sess.run(init)
 
     # Saver for storing checkpoints of the model.
-    saver = tf.train.Saver()
+    saver = tf.train.Saver(var_list=tf.trainable_variables())
 
     try:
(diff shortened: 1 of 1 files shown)


lemonzi:
Fair point. Some optimizers save an internal state that keeps track of the gradient distribution, I don't know what the effect would be (I guess not much). It would be also nice to save and restore the global step, but I think we're not doing that anyway.

ibab:
The fact that these tensors are stored is a bug IMO.

lemonzi:
Ok, let's merge this then. I also think the other frameworks such as tensorflow.slim only save trainable variables by default.

Issue #125, opened 2016-10-06
Master currently generates static


jyegerlehner:
I think I've found that the current master, which by default uses fast_generation, generates static. If fast_generation is disabled, then it seems to generate correctly. 

This conclusion is based on generating only two short samples. But I noticed the same behaviour in a development branch in which I had merged in master.

lelayf:
I'm making a similar observation. Last good master that is working for me with fast gen is 4dafacb.
Also i am now shuffling samples and wondering if that could impact the training process.

Zeta36:
**Yes!!** Fast generation is working badly.

I made the next experiment:

I modify the TextReader I made for the WaveNet text generation version (#117). specifically: I made a **ImageReader**. This reader just load (using Pillow: PIL) a Gif image, rescale it to 32x24 (pic.resize((32,24), Image.ANTIALIAS)) and finally get the 8bit grayscale channel (using convert('L')).

After all this, I push the (0-255) bytes from the image (grayscale) into a np vector (img = np.array(pic)). This is the result when I print it after reading a Gif image (32x24 matrix like this):
[[ 48  51  58  60  60  61  68  71  71  74  75  74  71  64  66  63  58  55
   53  58  56  55  54  60  62  63  65  70  73  70  63  46]
 [ 66  78  86  87  90  93  95 102 103  97  83  68  58  55  56  56  56  48
   44  46  47  50  66  85  91  91  93  95  99 100  91  62]
 [ 62  76  79  80  83  90  95  82  55  46  41  43  43  44  44  43  44  42
   39  35  33  37  45  54  65  79  91  91  92  94  90  61]
 [ 66  78  86  86  87  91  82  47  38  46  46  46  47  48  46  45  46  47
   45  46  49  44  39  44  47  57  73  89  97  99  92  62]
....
....
....
 [145 161 156 153 154 158 158 147 151 116 135 125 106 125 138 138 131 150
  145 129 160 134  90 180 219 208 203 202 220 223 214 186]
 [180 192 172 149 160 172 161 147 135 143 106 154 157 131 137 150 140 125
  121 172 187 116 152 204 196 200 203 211 219 217 210 180]]

I reshape this matrix later to one column with img = img.reshape(-1, 1)

Ok. Now I train the WaveNet model feeding with this 1D-image vector but using just a single image ("tp.gif") for all the training. The original (not scale) image is this (and once scaled its matrix is the one above): 
[...]

PR #126, opened 2016-10-07
Fixing the fast generation

Zeta36:
I think I found the problem with the fast generation. Problem probably was in a bug in this line in the _generator_dilation_layer() method:

conv_gate = output_gate + variables['gate_bias']

That it should be:

output_gate = output_gate + variables['gate_bias']

So, everyone that has used the bias parameter as True, has been generating noise for a long time. I test with the image generation (#125) and now all was fine with the fine fast generation (no different outputs between fast and standard generation).


Zeta36:
Ops! Please, do not merge the binary file I uploaded in my fork. Just the update in the model.py file  (53ad718).

nakosung:
I posted a *cleaned* version as #127.

ibab:
@Zeta36: Should I go ahead and merge #127 or do you want to remove the `.wav` file from this PR?

Zeta36:
@ibab I don't mind, you can go ahead with #127. Actually, I don't really know how to remove the .wav file from this pull request :P.

ibab:
Okay :)
Thanks a lot for fixing this!
I'm not sure why this problem didn't cause the unit tests to fail.

lemonzi:
There is a test for training with a sine wave and overfitting and a test
for generating white noise with an untrained net and doing size and range
checks. We should add a test that generates a sine wave with an overfitted
net to close the loop :)

El dv., 7 oct. 2016 a les 14:43, Igor Babuschkin (<notifications@github.com>)
va escriure:

jyegerlehner:
Thanks @Zeta36 for finding that and the fix! I confirm that it works.

Zeta36:
You are welcome @jyegerlehner :)

PR #128, opened 2016-10-07
Rmsprop optimizer

jyegerlehner:
Supercedes [PR 120](https://github.com/ibab/tensorflow-wavenet/pull/120).

### DIFF
--- a/test/test_model.py
+++ b/test/test_model.py
@@ -34,8 +34,8 @@ def MakeSineWaves():
 class TestNet(tf.test.TestCase):
     def setUp(self):
         self.net = WaveNetModel(batch_size=1,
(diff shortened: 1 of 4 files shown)


lemonzi:
I'd prefer a cleaner alternative to those globals, but we can merge it now and fix it later.

jyegerlehner:
Oops right, sorry. You commented about that and then I forgot.

Is this commit something like what you had in mind?

lemonzi:
Yes, this is great! Can you add it to the test as well?

jyegerlehner:
Just for you @lemonzi, it is done!

ibab:
This looks good. I'll go ahead and merge it.

PR #131, opened 2016-10-08
Added temperature flag to generation script

robinsloan:
It's nice to be able to specify sampling "temperature" when generating output, usually for aesthetic reasons, so I added some code to scale the sampling probabilities if a temperature other than 1.0 is provided.

Demo: https://soundcloud.com/robinsloan/sets/tensorflow-wavenet-temperature-demo

### DIFF
--- a/generate.py
+++ b/generate.py
@@ -7,7 +7,6 @@ import json
 import os
 
 import librosa
-
 import numpy as np
 import tensorflow as tf
 
@@ -30,9 +29,9 @@ def get_arguments():
         return {'true': True, 'false': False}[s.lower()]
 
     def _ensure_positive_float(f):
-        """Sampling temperature must be a positive float."""
+        """Ensure argument is a positive float."""
         if float(f) < 0:
-            raise argparse.ArgumentTypeError('Sampling temperature must be positive.')
+            raise argparse.ArgumentTypeError('Argument must be greater than zero')
         return float(f)
 
     parser = argparse.ArgumentParser(description='WaveNet generation script')
--- a/generate.py
+++ b/generate.py
@@ -7,6 +7,8 @@ import json
 import os
 
 import librosa
+import scipy.misc
+
 import numpy as np
 import tensorflow as tf
 
@@ -188,11 +190,9 @@ def main():
 
         # Scale sample distribution using temperature, if applicable.
         if (args.temp != 1.0 and args.temp > 0):
-            np.seterr(divide='ignore')
             prediction = np.log(prediction) / args.temp
-            prediction[np.isneginf(prediction)] = 0
-            prediction = np.exp(prediction) / np.sum(np.exp(prediction))
-            np.seterr(divide='warn')
+            prediction = prediction - scipy.misc.logsumexp(prediction)
+            prediction = np.exp(prediction)
 
         sample = np.random.choice(
             np.arange(quantization_channels), p=prediction)
--- a/generate.py
+++ b/generate.py
@@ -190,9 +190,11 @@ def main():
 
         # Scale sample distribution using temperature, if applicable.
         if (args.temp != 1.0 and args.temp > 0):
+            np.seterr(divide='ignore')
             prediction = np.log(prediction) / args.temp
             prediction = prediction - scipy.misc.logsumexp(prediction)
             prediction = np.exp(prediction)
+            np.seterr(divide='warn')
 
         sample = np.random.choice(
             np.arange(quantization_channels), p=prediction)
--- a/generate.py
+++ b/generate.py
@@ -13,6 +13,7 @@ import tensorflow as tf
 from wavenet import WaveNetModel, mu_law_decode, mu_law_encode, audio_reader
 
 SAMPLES = 16000
+TEMPERATURE = 1.0
 LOGDIR = './logdir'
 WINDOW = 8000
 WAVENET_PARAMS = './wavenet_params.json'
@@ -35,6 +36,11 @@ def get_arguments():
         type=int,
         default=SAMPLES,
         help='How many waveform samples to generate')
+    parser.add_argument(
+        '--temp',
+        type=float,
+        default=TEMPERATURE,
+        help='Sampling temperature')
     parser.add_argument(
         '--logdir',
         type=str,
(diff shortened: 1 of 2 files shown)


generate.py
```diff
@@ -179,6 +185,15 @@ def main():
 
         # Run the WaveNet to predict the next sample.
         prediction = sess.run(outputs, feed_dict={samples: window})[0]
+
+        # Scale sample distribution using temperature, if applicable.
+        if (args.temp != 1.0 and args.temp > 0):
+            np.seterr(divide='ignore')
+            prediction = np.log(prediction) / args.temp
+            prediction[np.isneginf(prediction)] = 0
+            prediction = np.exp(prediction) / np.sum(np.exp(prediction))
```
jyegerlehner: I think this line is better implemented as something like:
```
prediction = prediction - scipy.misc.logsumexp(prediction)
prediction = np.exp(prediction)
```
By operating in the log domain as much as possible we avoid divisions which can lead to instability when we divide by something very close to zero. Especially if we ever want to run using float16.

robinsloan: Never would have thought of that; thanks for the guidance!

robinsloan:
I made @jyegerlehner's suggested changes, which introduces a `scipy` dependency, which is… not great? But there's no direct `numpy` equivalent to the `scipy.misc.logsumexp` function, unfortunately.

jyegerlehner:
If we don't like the dependency, would logsumexp() be equivalent to np.log(np.sum(np.exp(prediction)))? 

I wouldn't uncritically paste my code in there; I was just describing an idea, and my code is not usually right the first time around :).

A test that shows that the temperature sampling with T=1 produces same distribution as without would be good. 

And perhaps, if I understand this, in the limit as T->0, it should be equivalent to choosing the most likely quantization level. And then as T-> inf, all quantization levels become equally likely?

robinsloan:
Great idea on the test. I've never written a Python test before so I'll poke at the existing tests in this project and figure it out.

I'll try that numpy approach. I tried something different (no.logaddexp.reduce) and didn't get what I expected so I think a more stepwise approach would would be better for me.

ibab:
I'm fine with using scipy as long as we put it into the `requirements.txt` :)
Note that `logsumexp(predictions)` is different from doing `log(sum(exp(predictions)))`.
It shifts by `max(predictions)` in order to avoid underflow.

This is a really nice addition to the project!

generate.py
```diff
@@ -179,6 +187,15 @@ def main():
 
         # Run the WaveNet to predict the next sample.
         prediction = sess.run(outputs, feed_dict={samples: window})[0]
+
+        # Scale sample distribution using temperature, if applicable.
+        if (args.temp != 1.0 and args.temp > 0):
```
lemonzi: Parentheses are not needed here.

generate.py
```diff
@@ -36,6 +39,11 @@ def _str_to_bool(s):
         default=SAMPLES,
         help='How many waveform samples to generate')
     parser.add_argument(
+        '--temp',
```
lemonzi: Please use `temperature` or `sampling_temperature`.

robinsloan:
Quick question for @jyegerlehner or anyone else as I'm fixing this up:

When I scale the prediction distribution with `temperature=1.0`, running it through log space, I get back the original distribution, as expected, though it's not identical -- it is `np.allclose` to within `1e-09`. Does that sound reasonable? (I have no intuition for this.)

jyegerlehner:
@robinsloan That sounds plenty good to me. We might want to future-proof for fp16, and make the tolerance wider, because float16 only has 6-7 significant digits. Maybe use 1e-4 for a tolerance?

robinsloan:
Updates:

* Application of temperature to prediction distribution now happens in log space
* A test ensures scaling at temperature=1.0 gives us back the original predictions, to within 1e-5
* No scipy dependency
* Style changes, per @lemonzi

Thanks for all the feedback & support on this!

generate.py
```diff
@@ -7,12 +7,14 @@
 import os
 
 import librosa
+
```
lemonzi: Why this blank line?

generate.py
```diff
@@ -179,8 +192,20 @@ def main():
...
+        # Scale prediction distribution using temperature.
+        np.seterr(divide='ignore')
+        scaled_prediction = np.log(prediction) / args.temperature
+        scaled_prediction = scaled_prediction - np.logaddexp.reduce(scaled_prediction)
+        scaled_prediction = np.exp(scaled_prediction)
+        np.seterr(divide='warn')
+
+        # Prediction distribution at temperature=1.0 should be unchanged after scaling.
+        if args.temperature == 1.0:
```
lemonzi: This should be a unit test, we can move it later when tests for generation are ready.

generate.py
```diff
@@ -27,6 +29,12 @@ def _str_to_bool(s):
                              'boolean, got {}'.format(s))
         return {'true': True, 'false': False}[s.lower()]
 
+    def _ensure_positive_float(f):
```
lemonzi: Shouldn't this be generic and not specific to the sampling temperature?

lemonzi:
For the other reviewers: these are minor nits, feel free to merge and mark to fix later.

generate.py
robinsloan: I didn't want to change anything outside the scope of this feature; maybe this suggestion is better left to a general refactoring of the argument parsing?

generate.py
robinsloan: I haven't written test coverage in Python before and couldn't determine where in the codebase to put this, if not here. Pointers appreciated!

lemonzi: You didn't find it because it doesn't exist yet -- there is a PR that provides tests for this file. Thanks for specifying how to test it, though! We'll move it later, no problem.

generate.py
robinsloan: Ohhh wait I see what you're saying. I didn't notice that the `ArgumentTypeError` indicated automatically the argument to which it was objecting. So the error can just say "yo you need a positive float." Got it got it.

lemonzi: Didn't you add this function as part of the feature? It has a very generic name but it's specific to the temperature parameter, that's why I mentioned this. I think `argparse` takes care of referring to the argument name when displaying an error, so I suggested leaving the "positive float" generic rather than adding a `_parse_temperature` function. The `_str_to_bool` is due for a refactor anyway though, so we could do both of them together later on, but I feel it's a but unnecessary.

lemonzi: Haha yep, we typed at the same time. Thanks!

lemonzi:
LGTM

jyegerlehner:
@lemonzi Would we want to wait to merge this until after PR 142 is merged? Those tests in 142 will be running this code, so we will get to see that they still pass before we merge.

ibab:
LGTM, too!
Feel free to merge.

lemonzi:
@jyegerlehner Good point, we can wait, there's only that conflict left to resolve.

ibab:
@jyegerlehner: Okay, that's a good idea. @robinsloan will have to rebase on top of `master` once we merge that PR so that travis will run your test.

jyegerlehner:
@robinsloan Could you push a dummy change to your branch to trigger another run of the tests, so we can see if they pass, now that the generation test has been merged? Thx

lemonzi:
(You can do that with `git commit --allow-empty -m "Trigger Travis"`)

robinsloan:
That's new to me—thanks @lemonzi!

PR #133, opened 2016-10-09
Delete unused arguments

jyegerlehner:
I started looking at speaker-id conditioning, and found that a number of arguments are passed in that aren't used in model.py.


ibab:
Yeah, thanks for cleaning these up!

ibab:
@jyegerlehner: I've made you a collaborator, so you'll be able to merge pull requests from now on.

PR #137, opened 2016-10-10
Replace sum() with tf.add_n()

jyegerlehner:
I think this might be better because on the GPU it would end up storing the result of the summation to global memory once, instead of n-1 times. I also fear a bunch of cascaded summations ends up creating multiple intermediate tensors which each have to be allocated and use up memory.

In other words, sum(A) creates a graph that does:
sum1 = a1+a2
sum2 = sum1 + a3
sum3 = sum2 + a4
...

instead of tf.add_n() which just does:
sum = a1 + a2 + a3 + a4 + ...

I don't really know for sure.

### DIFF
--- a/wavenet/model.py
+++ b/wavenet/model.py
@@ -325,7 +325,7 @@ class WaveNetModel(object):
 
             # We skip connections from the outputs of each layer, adding them
             # all up here.
-            total = sum(outputs)
+            total = tf.add_n(outputs)
             transformed1 = tf.nn.relu(total)
             conv1 = tf.nn.conv1d(transformed1, w1, stride=1, padding="SAME")
             if self.use_biases:
@@ -398,7 +398,7 @@ class WaveNetModel(object):
 
             # We skip connections from the outputs of each layer, adding them
             # all up here.
-            total = sum(outputs)
+            total = tf.add_n(outputs)
             transformed1 = tf.nn.relu(total)
 
             conv1 = tf.matmul(transformed1, w1[0, :, :])


ibab:
It actually used to be `tf.add_n` in one of the earlier commits :)
I think `sum` might be better because TensorFlow can clean up  the `a1`, `a2`, ... tensors as it calculates each of the residual blocks. So I would expect that it uses less memory.
We should use the memory profiling functionality in TensorFlow to find out.

jyegerlehner:
> So I would expect that it uses less memory.

That could be true, but still be slower, as it requires more writes to memory compared to just storing the final sum once. 

> We should use the memory profiling functionality in TensorFlow to find out.

Any pointers to that? I haven't found any memory profiling in tensorflow docs.

ibab:
> Any pointers to that? I haven't found any memory profiling in tensorflow docs.

If you enable the `store_metadata` option, the training script will save the profiling information in the TensorBoard output: https://github.com/ibab/tensorflow-wavenet/blob/2125f3966e38ebaa8109241c27c4e4c1854f9ace/train.py#L272
You might have to add the CUDA profiling libraries to your LD_LIBRARY_PATH.

In the graph tab in TensorFlow, you'll be able to pick the log output that contains the profiling information on the left, which will give you memory and evaluation statistics.

There's also a timeline trace that you can open with the Chrome timeline viewer.

lemonzi:
This appeared in a PR but was removed because it was an unrelated change, it'd be good to discuss it a bit further here. I think that if TensorFlow provides something natively that can be so trivially done in Python there has to be a good reason for it. However, one of the Google Brain engineers pointed out that adding them manually indeed uses less memory for large tensors (see http://stackoverflow.com/a/34520066/1729279). Why do they provide `add_n` then? To simplify the graph?

jyegerlehner:
> Why do they provide add_n then?

I tried to say above: I imagine it would reduce from N stores to global memory to 1. Faster. 

I don't guess the memory thing being significant. I don't know the intricacies of tf's heap coalescing and temporary tensor allocations, but it seems like it would only slightly change how long it has to keep the operand tensors in memory. And I wonder, if they are being added to summaries as we are doing, if it can even let go of them at all.

> If you enable the store_metadata option, the training script will save the profiling information in the TensorBoard output

Thanks for that. There's a list of things I'd like to get done and right now this doesn't seem like my best use of time, so don't think I will work on that.

Issue #141, opened 2016-10-11
Model fails to learn sine wave


atgambardella:
With the default parameters (simply cloning the repo and training), the model could not fit a sine wave (I was intentionally trying to overfit on one of the simplest examples as a sanity check).

The model seems to converge at a loss of 0.001~0.002. The input and output are here: [https://soundcloud.com/andrew-gambardella-769364795/sets/sine](https://soundcloud.com/andrew-gambardella-769364795/sets/sine)

Notice that the input sine wave sounds a lot cleaner than the output, and that the generated waveform as shown by soundcloud looks completely flat, compared to the bumpy input sine wave.

I think it would be useful to get this working before moving onto more complicated examples such as voice.

atgambardella:
Here are the two waveforms plotted (after appropriate rescaling etc.): http://imgur.com/a/sej9i

Notice that around the peaks, the generated waveform isn't smooth, which I think is what's causing the unclean sound.

Zeta36:
@atgambardella, try changing this line in generate.py (line 182):

`sample = np.random.choice(np.arange(quantization_channels), p=prediction)`

by this:

`sample = np.argmax(prediction)
`
And tell me. 

If it doesn't work, change too the line (148):

`waveform = np.random.randint(quantization_channels, size=(1,)).tolist()`

by the first float value of the sine wave. For example:

`waveform = [.5]`

I used this method with a overfitted model and I could replicate the original sample. I did it for example in here #125 and in here #129 (I use a image raw input but it the same)

Regards.

atgambardella:
@Zeta36 I tried your first suggestion - replacing the sample with argmax in the generator - and I'm still getting the same non-smooth behavior at the peaks.

Your second suggestion would align them, but seeding the generator wouldn't change that behavior at the peaks, correct?

My input sine wave is sampled at 44.1kHz -- could this be part of the issue?

Zeta36:
In my tests, just with argmax and seeding the first value of the waveform both at same time was the only way I could have a perfect generation in a overffited model. 

Try figure out the first value of that sine wave and seed as I told you the wavwform. I think you will have a smooth output in that way.

lemonzi:
The default settings now use a larger net and train for longer. The former ones were for a toy network. Can you test again to see if it performs better?

atgambardella:
[...]

PR #142, opened 2016-10-11
Add generation test

jyegerlehner:
I hacked in a test to check generation. Right now it just tests WaveNetModel.predict_proba(), not yet WaveNetModel.predict_prob_incremental(). I duplicated a bit of code from generate.py to do it; it's a bit ugly.

The only idea I came up with was to check that the power spectrum of the generated waveform matches what we trained it on. So the training signal looks like this:

sine_train

with peaks at the three frequencies we expect.

A successfully generated waveform looks like this:
sine_generated

So the test just adds up the power in generated waveform's spectrum, and asserts that at least 90% of it is at those three frequencies. 

I had to train longer to get good results, and generation takes a while. I'm afraid our CI test is getting prohibitively-long. Takes a couple minutes even on the GPU. Afraid to see how long it takes on the CI server.

Is this reasonable approach? Suggestions?

And just for fun, [this](https://soundcloud.com/user-731806733/tf-wavenet-unit-test-sinewave) is what our training signal sounds like.

### DIFF
--- a/test/test_model.py
+++ b/test/test_model.py
@@ -99,7 +99,7 @@ def check_waveform(assertion, generated_waveform):
     f3_power = find_nearest(freqs, power_spectrum, F3)
     expected_power = f1_power + f2_power + f3_power
     # print("Power sum {}, F1 power:{}, F2 power:{}, F3 power:{}".
-    #        format(power_sum, f1_power, f2_power, f3_power))
+    #       format(power_sum, f1_power, f2_power, f3_power))
 
     # Expect most of the power to be at the 3 frequencies we trained
     # on.
--- a/test/test_model.py
+++ b/test/test_model.py
@@ -99,7 +99,7 @@ def check_waveform(assertion, generated_waveform):
     f3_power = find_nearest(freqs, power_spectrum, F3)
     expected_power = f1_power + f2_power + f3_power
     # print("Power sum {}, F1 power:{}, F2 power:{}, F3 power:{}".
-    #       format(power_sum, f1_power, f2_power, f3_power))
+    #        format(power_sum, f1_power, f2_power, f3_power))
 
     # Expect most of the power to be at the 3 frequencies we trained
     # on.
@@ -186,6 +186,7 @@ class TestNet(tf.test.TestCase):
 
 
 class TestNetWithBiases(TestNet):
+
     def setUp(self):
         self.net = WaveNetModel(batch_size=1,
                                 dilations=[1, 2, 4, 8, 16, 32, 64,
@@ -233,8 +234,8 @@ class TestNetWithScalarInput(TestNet):
                                 skip_channels=32,
                                 scalar_input=True,
                                 initial_filter_width=4)
-        self.optimizer_type = 'sgd'
-        self.learning_rate = 0.02
+        self.optimizer_type = 'rmsprop'
+        self.learning_rate = 0.001
         self.generate = False
         self.momentum = MOMENTUM_SCALAR_INPUT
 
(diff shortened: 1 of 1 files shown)


ibab:
This is really clever.
Maybe we can get it to converge faster by increasing the learning rate.

jyegerlehner:
I spent a fair bit of time tweaking learning rates, sizes of the dilation stacks, momentum, optimizer & cetera trying to reduce training time. Which is not to say what you suggest isn't possible. But the CI server tests only took about 16 minutes. I don't see that as prohibitive. It's not as bad as I feared.

jyegerlehner:
I added testing of fast generation. Cleaned up a smidge. I think something that doesn't duplicate some code from generate.py would require some refactoring of code in generate.py that can be called from test_model.py.

I didn't hear any howls of protest, so I'll remove WIP.

lemonzi:
Setting the right frequencies for the sinusoids can get rid of the leak to neighboring bins, and we could then tighten that threshold. I believe the frequencies must be set so that the length of the signal is a multiple of the period.

lemonzi:
So, `f_n = n / SAMPLE_DURATION`, and no need to use the sampling rate, just use `1, 2, 3, ...` as `time`. And I'd suggest using `cos` for all of them, it's more standard (the phase in the FFT is then 0 for all sinusoids). We can also change the relative phase in some of them in a more difficult test case.

lemonzi:
We could maybe move the synthesis and FFT-based evaluation to a separate file, so that `test_model.py` only tests the network. It's growing a bit long. We could also have a `test_model_base.py`, and then define all the subclasses (biases, different settings, etc) in a separate file.

lemonzi:
As mentioned in #141 sometimes the error is small but it's there, so we need to be strict. 

I would also test using the power in dB, rather than in linear units. In that case, rather than squaring it just use `20 * np.log10(np.abs...`

jyegerlehner:
> As mentioned in #141 sometimes the error is small but it's there, so we need to be strict. 

I got the impression in 141 that training may have just stopped at 4K iterations, as he said he used defaults. So it wasn't perfect yet. The wave file posted sounded to me like it learned the waveform. I don't agree that 141 identifies any kind of bug that this test should detect. This test was originally designed as a smoke test that we can create a model, train, and learn. You pointed out it we needed to close-the-loop by checking that we generate from the trained model. With this PR, it does that too. 

Since the purpose is to check for bugs, not to demonstrate hyper-fidelity in a trained model, I think we do not want to be "strict". 

As for your other refinements, can those not come along in a future refactoring, when someone thinks they are the best use of his time? They sound like a large investment of time for possibly-arguable stylistic benefit. I'd rather have slightly less-than-optimal-test than no test at all.

lemonzi:
Yes, I agree it's better to have this as soon as possible to increase code coverage in the tests. Let's merge it as-is. Can you fix the conflict with the RMSProp?

jyegerlehner:
Oops I didn't notice the conflicting file; thanks for the pointer.

jyegerlehner:
Oops wait, I just fixed the conflicts. Must be from that last merge.

lemonzi:
Still doesn't like it 😮

lemonzi:
Maybe you need to pull from this repo onto your fork's master and then rebase, I looked at the network graph and you're a commit behind. That commit was yours, but since it re-writes them maybe it didn't pick it up -- or you don't have it merged on your fork..

jyegerlehner:
Thanks for the pointers. This stuff is too complicated for my simple mind. I'll work it out eventually.

ibab:
Okay, this looks good :)

Issue #136, opened 2016-10-10
can't get loss below 5.545


dlahoti:
For some reason, I can't get a loss below 5.545¸ regardless of options or dataset. Generating with the trained models just gives me static.

jyegerlehner:
If you're using defaults, you need to decrease the learning rate to 0.001 just to get it learning something. I think one also needs larger receptive field than what's in the default wavenet_params.json. 

There're example params files over in the issue thread on "Generating good audio samples".

Note to self: PR to change those defaults.

ibab:
Yeah, we should definitely change the broken defaults :)

ibab:
We have better defaults now, and @jyegerlehner also fixed a different issue that was causing instability in the optimization, so I would consider this fixed now.

PR #145, opened 2016-10-12
Better default values

jyegerlehner:
We've had a couple of issues where people naively take the default values and try training and don't get good results. So here are defaults I've found work. Actually I'd rather choose a dilation stack with a 500 msec receptive field, and increase the number of skip_channels, but fear that might not fit on your typical 8 GB GPU. 

Any objections?


lemonzi:
I'd prefer "nice" numbers for a default -- how does it behave at step = 100.000?

I usually specify the steps in exponential notation, as `int(1e5)`, and same for the learning rate, `1e-3`, as you usually change orders of magnitude to see how it performs before doing fine-tuning. I'm good with what you have now if you prefer it, though.

lemonzi:
Closes #136.

jyegerlehner:
> I'd prefer "nice" numbers for a default -- how does it behave at step = 100,000?

I don't know; I just added a couple of zeros. I get the impression that some people think that it's "done" when the arbitrarily-chosen 4000 steps is hit. And if the trained model isn't perfect at that point, then it's a bug in the library. I just wanted to make it run "forever", so it might dawn on the naive user that he has to make a decision when it's trained long enough. Yes, 1e5 is a nicer number.

jyegerlehner:
>  how does it behave at step = 100,000

We can probably say poorly. If only because the default scenario is training on the whole corpus, and without speaker_id global conditioning and speaker shuffling I think it mostly forgets the previous speaker's voice as it mows down 400 files in a row from the next speaker.

jyegerlehner:
Thumbs up or down to vote please.

ibab:
It seems that everyone wants this to get merged :)
One more thing that I just noticed: Should we also switch to scalar input at the same time?
I have a mail from one of the DeepMind authors saying that they've fed in the waveform directly as the input (no one-hot encoding).
I think it doesn't matter in terms of the quality of the output, but it increases efficiency as we don't have to deal with the `[filter_width, quantization_channels, residual_channels]` convolution.

jyegerlehner:
> Should we also switch to scalar input at the same time?

Last I checked we can't do fast generation with scalar inputs. Also, in my experiments I didn't get it to work as well as the quantized input; or at least training loss was quite a bit higher, ceteris paribus. I haven't tried generating with it yet, now that I think about it.  I figure I'd have another go at it here before long, perhaps try training with rms prop. At this point, I don't think I'd make it the default.

ibab:
Ah, I forgot about the fast generation.
The changes required for that shouldn't be too difficult,
but we should merge this PR now, and think about scalar input later.

Issue #143, opened 2016-10-11
Strange growth in the loss


Zeta36:
This is the case:

I start to train a little dataset so I can get a very small (overfitted) loss. In the beginning all goes well. I start with a ~5.5 error cost, and while training the loss drop gradually until I reach a near ~0.05 loss. but, and this is the question: if I continue training after this low loss has been reached; after some hundred steps more, the cost begins to **grow** instead of staying near 0!! (sometimes, the loss go up to a value much above 5??)

I don't know what can cause this strange behavior, but I guess it's not a normal thing.

What do you think?

Zeta36:
May be the case that simply the Reader get out of data to feed into the sample_placeholder (as I use a little data set)?

ibab:
@Zeta36: From what I've seen, the problem is that the ReLU layers at the end cut off the entire network output. The following things seem to help:
 - use a smaller learning rate
 - use regularization
 - replace the ReLU layers with something that's nonzero almost everywhere like Tanh

jyegerlehner:
> From what I've seen, the problem is that the ReLU layers at the end cut off the entire network output.

If that's what's happening, I wonder if an option to make the "post-processing" ReLU layers res blocks would help. Residual blocks usually make things better anyway, allowing activations and gradients to go around the layer.

jyegerlehner:
@Zeta36 @ibab 

**TL;DR:** I think I may have a fix for what Zeta36 describes. They are in [PR 128](https://github.com/ibab/tensorflow-wavenet/pull/128) for rmsprop and [PR 147](https://github.com/ibab/tensorflow-wavenet/pull/147) for adam.

### Explanation

I've seen the behavior Zeta36 is describing in our test/test_model.py. When that test uses adam or rmsprop, I would see the loss drop and drop till a small number, then jump up to a large loss at some random time. 

You can reproduce that problematic behavior if you change (what at the time of this writing in master) [MomentumOptimizer  to AdamOptimizer](https://github.com/ibab/tensorflow-wavenet/blob/master/test/test_model.py#L58), and make the learning rate 0.002. Uncomment the statements that [print loss](https://github.com/ibab/tensorflow-wavenet/blob/master/test/test_model.py#L73).

If you run the test with 
```
python test/test_model.py
```
[...]

PR #147, opened 2016-10-13
Make adam more stable

jyegerlehner:
This is on top of the RMS-prop PR. The changes unique to this PR start at the comment "Make adam more stable". 

I describe the rationale for this change in the [issue 143](https://github.com/ibab/tensorflow-wavenet/issues/143) thread.

### DIFF
--- a/test/test_model.py
+++ b/test/test_model.py
@@ -41,8 +41,8 @@ class TestNet(tf.test.TestCase):
                                 dilation_channels=32,
                                 quantization_channels=256,
                                 skip_channels=32)
-        self.optimizer_type = 'sgd'
-        self.learning_rate = 0.02
+        self.optimizer_type = 'adam'
+        self.learning_rate = 0.002
 
     # Train a net on a short clip of 3 sine waves superimposed
     # (an e-flat chord).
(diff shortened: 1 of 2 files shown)


ibab:
This is excellent. I didn't even know about the `epsilon` parameter.
Unfortunately we can't merge as I've squashed your previous PR before merging it into master.
You can update your PR like this (no need to make a new one):
```
# Make sure that we have the newest changes on master
git checkout master
git pull
# Make a new branch
git checkout -b make-adam-stable-2
# Add the last commit from this PR
git cherry-pick 98783e5
# Overwrite this PR by running
git push -f origin make-adam-stable-2:make-adam-stable
```

nakosung:
I think it would be much safer if we amplify loss instead of decreased
epsilon. (Fp16)

2016년 10월 14일 (금) 오전 3:29, Igor Babuschkin <notifications@github.com>님이 작성:

jyegerlehner:
@nakosung  I don't think I understand what you are proposing. Could you elaborate a bit please? I think we increased, not decreased, epsilon.

nakosung:
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/training/adam.py#L63
says it is not good default value in general. I didn't read the code and
thought it would be safe to feed trainer with healthy range but I think I
was wrong. ;)

2016. 10. 14. 오전 9:33에 "jyegerlehner" <notifications@github.com>님이 작성:

jyegerlehner:
Nice find on comment about epsilon. I missed that. It's surprising they buried it there, and don't mention it where everyone would see it (in the API docs).

ibab:
This looks good, merging now.

PR #151, opened 2016-10-15
Exit without error when num_steps exceeded

jyegerlehner:
There is a defect where, when you resume training from a checkpoint and the saved num steps already exceeds the --num_steps, it exits with a python exception saying that step is referenced before assignment. This is a trivial fix for that.


ibab:
Nice catch!

PR #152, opened 2016-10-17
Check for existence of audio files

### LINKED ISSUE #150: python train.py hangs on wrong path instead of raising error
Hi all, awesome project! I checked it out and wasn't able to run the training procedure.

Running 
`python train.py ` causes the code to run properly up to this line:
`               summary, loss_value, _ = sess.run(
                    [summaries, loss, optim],`
At this point the code hangs forever.

I think this may be a symptom of the queue runners failing to fetch input data, I got the same error when I don't have a dataset file. However, I have downloaded the VCTK dataset and placed it in the correct folder. Any help in figuring out how to debug the data fetching part (or otherwise resolving this issue) would be greatly appreciated.

If it matters, I'm running tensorflow '0.11.0rc0'

ibab:
This fixes #150.

### DIFF
--- a/wavenet/audio_reader.py
+++ b/wavenet/audio_reader.py
@@ -72,6 +72,12 @@ class AudioReader(object):
                                          shapes=[(None, 1)])
         self.enqueue = self.queue.enqueue([self.sample_placeholder])
 
+        # TODO Find a better way to check this.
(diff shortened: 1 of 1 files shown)


jyegerlehner:
LGTM

PR #157, opened 2016-10-21
Make summary histograms optional

jyegerlehner:


lemonzi:
Great! This was long overdue 😄 

What about the loss summaries? It probably makes sense to only drop the histograms, as you suggest. In that case, we should rename the flag to something like `histogram_summaries` or `add_histogram_summaries`. 

I would maybe make it false by default (it's only useful when debugging).

jyegerlehner:
> What about the loss summaries?

I don't see much need; loss makes a trivial contribution to the even log file from what I can tell. Mainly I think we want to be able to shut off the histograms.

> we should rename the flag to something like histogram_summaries

Ah, of course. THanks. I have trouble with lucid naming.

> I would maybe make it false by default (it's only useful when debugging).

Good point. I usually try to make things default to the way they behave before the change so that people don't get sudden unexpected changes in behavior. But disk space used by the event log is getting exorbitant for long-running training. And I don't think I've ever even looked at the histograms.

lemonzi:
The code looks good, but the tests are still running... Any ideas? If they got stuck, we can try re-triggering them.

jyegerlehner:
Actually the tests are not running. If you click on details it shows they passed.

I just think Github's icon is broken.

lemonzi:
Oh, I see... it may have to do with that downtime they had because of the DDoS. LGTM, I'll merge when it finishes.

Issue #129, opened 2016-10-08
WaveNet implementation for image generation (based in this repository) 


Zeta36:
As I told in #125, I've made a revision of this project where a feed the model using the gray-scale channel of a dataset of images (similar **but no equal** to what DeepMind made with **PixelCNN**).

I modify the TextReader I made for the WaveNet text generation version (#117). Specifically: I made a ImageReader. So this reader now can load (using Pillow: PIL) image files, rescale them to for example 64x64 (because I have no GPU and no time to waste :P) (pic.resize((32,24), Image.ANTIALIAS)) and finally get the 8bit grayscale channel (using convert('L')).

I really love the theoretical idea after the generative model using dilated causal convolution layers!! The results are wonderful and easily extrapolated.

I made the next test with this image implementation: 

1) I used a small dataset with 12 similar (but different) images of a person (the woman in the picture is my mother). She is posing in different ways.
![20140211_202000](https://cloud.githubusercontent.com/assets/17341905/19213557/c6087b66-8d66-11e6-9ba1-93bd6bfd337c.jpg)

Above one of the original pictures, and below the scaled gray-scale image fed into the training time:
![example_train_file](https://cloud.githubusercontent.com/assets/17341905/19213573/09cfec26-8d67-11e6-9f35-22e49e62eccd.jpg)

2) I trained the model using a SAMPLE_SIZE = 4096 and a learning rate of 0.001

3) The loss dropped down gradually and only after 4k steps it was around ~0.09. I trained once again with learning rate = 0.0001, and after another 4k steps loss was around ~0.01

4) I generated the images using always a WINDOW = 4096 and a always 4096 samples (64x64 pixels so I can save a full image).

Well:

5) If a let the first sample as random: 
`waveform = np.random.randint(quantization_channels, size=(1,)).tolist()`
or if I let the image grow using probability distribution and not argmax:
`sample = np.random.choice(np.arange(quantization_channels), p=prediction)`

I juts get ugly noise:
![mam3000_06](https://cloud.githubusercontent.com/assets/17341905/19213581/52df12d4-8d67-11e6-8225-d01399b3f717.gif)
![mam3500_02](https://cloud.githubusercontent.com/assets/17341905/19213582/591745e0-8d67-11e6-8516-f4bd05bec6d4.gif)

But, and this is interesting:

6) If I set the first sample to the first pixel in each image from the dataset. For example:
`waveform = [169]`
and if I use argmax to select every sample:
[...]

PR #162, opened 2016-10-25
Add related projects to the readme.

### LINKED ISSUE #117: My implementation of WaveNet for text generation (based in this repository)
Hi, friends.

As I have not a good GPU for heping you directly in this, I have use the baseline of the work in this repository to develop a WaveNet text generator (self-generator): [https://github.com/Zeta36/tensorflow-tex-wavenet](https://github.com/Zeta36/tensorflow-tex-wavenet).

In summary: I utilize the WaveNet model as a **text generator**. I feed the model using a raw text data (characters), instead of raw audio files, and once the network is trained, I use the conditional probability found to generate samples (characters) into an self-generating process.

Only printable ASCII characters (Dec. 0 up to 255) is supported right now.
Results

And pretty interesting results are reached!! Feeding the network with enough text and training, the model is able to memorize the probability of the characters disposition (in a lenguage), and generate later even a very similar text!!

For example, using the Penn Tree Bank (PTB) dataset, and only after 15000 steps of training (with low set of parameters setting) this was the self-generated output (the final loss was around 1.1):

"Prediction is: 300-servenns on the divide mushin attore and operations losers nis called him for investment it was with as pursicularly federal and sotheby d. reported firsts truckhe of the guarantees as paining at the available ransions i 'm new york for basicane as a facerement of its a set to the u.s. spected on install death about in the little there have a $ N million or N N bilot in closing is of a trading a congress of society or N cents for policy half feeling the does n't people of general and the crafted ended yesterday still also arjas trading an effectors that a can singaes about N bound who that mestituty was below for which unrecontimer 's have day simple d. frisons already earnings on the annual says had minority four-$ N sance for an advised in reclution by from $ N million morris selpiculations the not year break government these up why thief east down for his hobses weakness as equiped also plan amr. him loss appealle they operation after and the monthly spendings soa $ N million from cansident third-quarter loan was N pressure of new and the intended up he header because in luly of tept. N million crowd up lowers were to passed N while provision according to and canada said the 1980s defense reporters who west scheduled is a volume at broke also and national leader than N years on the sharing N million pro-m was our american piconmentalist profited himses but the measures from N in N N of social only announcistoner corp. say to average u.j. dey said he crew is vice phick-bar creating the drives will shares of customer with welm reporters involved in the continues after power good operationed retain medhay as the end consumer whitecs of the national inc. closed N million advanc"

This is really wonderful!! We can see that the original WaveNet model has a great capacity to learn and save long codified text information inside its nodes (and not only audio or image information). This "text generator" WaveNet was able to learn how to write English **words and phrases** just by predicting characters **one by one**, and sometimes was able even to learn what word to use based on context.

This output is far to be perfect, but It was trained in a only CPU machine (without GPU) using a low set of parameters configuration in just two hours!! I hope somebody with a better computer can explore the potential of this implementation.

You can download the new development in here: [https://github.com/Zeta36/tensorflow-tex-wavenet](https://github.com/Zeta36/tensorflow-tex-wavenet).

Technically:

1) I made a TextReader for feeding and replace the AudioReader.
2) I used the printable character ASCII decimal value (0-255) as the 8bit sample (and I remove the mu_law function from everywhere).
3) Removed all TensorBoard summaries (I have no memory to waste :P).
4) Removed wite_wav() and developed a write_text()
5) Some other minor changes: I start the "waveform" always with a space (char 32) and not with a random int, changed some terminal arguments, etc.

And that all!! 

I hope this can help you in any way.

Best regards,
Samu.

lemonzi:
This acknowledges related forks that diverge too much to be merged.

We could also merge the changes, but I feel by the time the refactors are complete it will be easier to just re-implement it.

With this I would close #117 and #129.

If you want to keep the issues open instead for discussion that's fine too, but they don't seem very active anymore.

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,7 +115,6 @@ To disable fast generation:
 python generate.py --samples 16000 model.ckpt-1000 --fast_generation=false
 ```
 
-
 ## Running tests
 
 Install the test requirements
@@ -132,3 +131,8 @@ Run the test suite
 
 Currently, there is no conditioning on extra information like the speaker ID.
 
+
+## Related projects
+
+- [tex-wavenet](https://github.com/Zeta36/tensorflow-tex-wavenet), a WaveNet for text generation.
+- [image-wavenet](https://github.com/Zeta36/tensorflow-image-wavenet), a WaveNet for image generation.


lemonzi:
I see now that @ibab was talking about merging it already and creating a TextReader (at least for the text one, which seemed to be most promising). That would also work.

Zeta36:
@lemonzi , I'm merging right now the TextReader and the ImageReader from #117 and #129. In some hours I'll send a PR with the changes. I hope this can help.

PR #165, opened 2016-10-26
fix a bug of wrong dimension when sample_size=0

fehiepsi:
When we want to feed all the audio file into the queue with sample_size=0, after `trim_silence()`, the `audio` variable is a row vector, which should be reshaped before `sess.run(self.enqueue, feed_dict={self.sample_placeholder: audio})`.

### DIFF
--- a/wavenet/audio_reader.py
+++ b/wavenet/audio_reader.py
@@ -95,6 +95,7 @@ class AudioReader(object):
                 if self.silence_threshold is not None:
                     # Remove silence
                     audio = trim_silence(audio[:, 0], self.silence_threshold)
+                    audio = audio.reshape(-1, 1)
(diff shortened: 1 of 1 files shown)


jyegerlehner:
Thank you for that @fehiepsi 

I reproduced the problem and verified the fix. Merging.

lemonzi:
@jyegerlehner For future PR's, could you "Squash and merge"? It leaves a much cleaner commit history (a single commit by the author at the time of merging and with you as a merger rather than a separate merge commit and the commit by the author in the past). It's also the default option for me, maybe it remembers whatever you did last.

jyegerlehner:
@lemonzi oops sorry. I guess I wasn't paying attention to that. Thanks for the pointer.

Issue #112, opened 2016-09-29
Global condition vs. Local conditioning


thomasmurphycodes:
In the white paper, they mention conditioning to a particular speaker as an input they condition globally, and the TTS component as an up-sampled (deconvolution) conditioned locally. For the latter, they also mention that they tried just repeating the values, but found it worked less well than doing the deconvolutions.

Is there effort underway to implement either of these? Practically speaking, implementing the local conditioning would allow us to begin to have this implementation speak recognizable words.

ibab:
Yeah, it's definitely a planned feature.
I'll get to it eventually, but I'd also accept contributions if someone is interested.
A solution to this should also integrate with the `AudioReader` interface.

Zeta36:
Is somebody working on this already?

alexbeloi:
I'm starting to work on it, I think I can get some basic implementation working over the next couple of days. Global part should be easy, and a dumb implementation (upsampling by repeating values) of local conditioning should be fast to implement as well.

This way, we can get to a stage where the net can produces some low-quality speech. Then work on improving the quality by adding more sophisticated upsampling methods.

The white paper also talks about using local conditioning features beyond just the text data, they do some preprocessing to compute phonetic features from the text. That would be nice to add later as well.

thomasmurphycodes:
I agree global will be easier, should just be a one-hot vector representing
the speaker. Am I thinking about this wrong that the local conditioning
requires us to train on data sets that contain the phonetic data as a
feature vector in addition to the waveform feature? What dataset are you
thinking of using?

alexbeloi:
I was thinking to just use the raw text from the corpus data for local conditioning to start, just encode each character into a vector and upsample (by repeats) it to the number of samples in the audio file, not ideal but it's a start. Characters should be able to act as a really rough proxy for phonetic features.

Ideally, the raw text should be processed (perhaps via some other model) into a sequence of phonetic features and then *that* would be upsampled to the size of the audio sample.

thomasmurphycodes:
I mean let's give it a shot and see what happens. Google Research has a
bunch of papers over on there page about HMM-ing characters to phonemes, so
we could look into a subproject where we try to implement that.

nakosung:
@thomasmurphycodes Could you post the list of papers?

thomasmurphycodes:
Yeah will tomorrow when in the office, they're on a box I have there.

ibab:
[...]

PR #168, opened 2016-11-01
Global conditioning

jyegerlehner:
There is a unit test in test_model.py. 

This code doesn't provider an AudioReader does not send back the speaker id, so it's not ready for use quite yet.

There are other global conditioning implementations in flight. Related discussions [here](https://github.com/ibab/tensorflow-wavenet/issues/112).  Let's try to find the most expeditious way of getting something merged from the various implementations. We're past-due for it IMO.


jyegerlehner:
How about that. We get a different numerical result running under python 3.5 than python 2.7.

Guess I'm going to have to figure out how to to run all this locally under python 3.5.

jyegerlehner:
I didn't hear anything from our friends alexbeloi or Zeta36 regarding their versions of global conditioning so I've carried on, when I've had time.

So it's at the stage where I've got it to work. I trained on the full corpus.

When I told it to generate for speaker id 311, I  got [this](https://soundcloud.com/user-731806733/generated-n-p311).

When I told it to generate for speaker id 245, I got  [this](https://soundcloud.com/user-731806733/generated-n-p245).

When I told it to generate for speaker id 233, I got [this](https://soundcloud.com/user-731806733/generated-n-p233).

So to me subjectively, if you listen to samples of those speaker ids in the vctk corpus, the generated samples sound about right. The quality of the samples is no better than before when I trained on just a single person, which is disappointing. I find when I train longer the loss goes down but the samples start sounding worse. So I imagine there's overfitting going on. So I think we need to separate the data into training and test sets, so we can watch for the test loss starting to rise, which I expect to see even while training loss is falling. I think I'd like to work on that next. Or other ideas?

At this point I think it's ready for review. My main reservations are: 1) I someone else needs to independently test this. Perhaps follow the instructions I put in the [README.md](https://github.com/jyegerlehner/tensorflow-wavenet/blob/global-conditioning/README.md) and see if it works, and check if we've broken things for the non-globally-conditioned case. I think it should still train and generate from existing models. 2) The test_model.py code has gotten uglier with `if global_conditioning: blah blah else blah blah`, more so than it was when @lemonzi was suggesting we refactor that. I think my preference might be to do that in a separate PR, but then it's also bad practice to allow the accumulation of technical debt. I guess I need a nudge in one direction or the other.  3) Code review please? 

So call for volunteers from amongst those who would like to see this merged. @ibab ? @lemonzi ? Bueller? Anyone?

Zeta36:
@jyegerlehner, you are doing a great job with the global conditioning. I don't have too much time right now to spend in this project but I'm glad to see you are going forward. I think you could open a issue for people testing your advances (I don't have a good GPU).

Keep it up!!

fehiepsi:
I will take part in reviewing this pull request. Will give a feedback in a couple of days.

mortont:
Just saw this PR, thanks for getting this done! I'm training a model with it right now and will get back to you once I have results.

Those generated samples sound good, I'm also thinking a more quantitative way to rate results would be to compare the fundamental frequency of a generated sample with that of the speaker's training samples. I don't have a very good ear for tone, so this may eliminate the "sounds about right" analysis. I'll see what I can do when this model's done.

I think it'd be fine to refactor `test_model.py` in another PR, but that's just my opinion. I can also give another set of eyes to a code review if needed.

wavenet/audio_reader.py
```diff
@@ -8,6 +9,27 @@
...
 
 
+def get_category_cardinality(files):
+    id_reg_expression = re.compile(r'p([0-9]+)_([0-9]+)\.wav')
+    min_id = None
+    max_id = None
+    for filename in files:
+        matches = id_reg_expression.findall(filename)[0]
+        id, recording_id = [int(id_) for id_ in matches]
+        if id < min_id or min_id is None:
```
mortont: Testing with python 3.4.2 this should be `if min_id is None or id < min_id:` otherwise it will throw `TypeError: unorderable types: int() < NoneType()`

wavenet/audio_reader.py
```diff
@@ -8,6 +9,27 @@
...
+            min_id = id
+        elif id > max_id or max_id is None:
```
mortont: Same reason as above, `elif max_id is None or id < max_id:`

fehiepsi: I can confirm this and the above.

jyegerlehner:
Thanks for the comments so far.

> a more quantitative way to rate results would be to compare the fundamental frequency of a generated sample with that of the speaker's training samples. I don't have a very good ear for tone, so this may eliminate the "sounds about right" analysis.

Hah hah, that sounds good. As far as "sounds about right", my observations were merely that 233 is female, and 311 and 245 are male. And 311 has a deeper voice than 245. All of which was preserved in the generated samples. Let us know what your fundamental frequency analysis shows.

fehiepsi:
Hi @jyegerlehner , I am still training right now. According to my calculation, our data has 100 persons, with the default setting, we have `sample_size=100000` with `sample_rate=16000`, so `1 sample_size ~ 6s data`. Each person has 150MB data which I expect `1s ~ 100kB`, so we have `1500s` for each person, which means our data contains around `150,000s` (including silence before and after the sound in a file). So it will take around 25000 iterations to go through the whole corpus. Am I right? How many steps have you run to get your result? (based on these information, we can get a little bit about which causes overfitting).

wavenet/audio_reader.py
```diff
@@ -105,14 +169,22 @@ def thread_main(self, sess):
                 if self.sample_size:
                     # Cut samples into fixed size pieces
                     buffer_ = np.append(buffer_, audio)
-                    while len(buffer_) > self.sample_size:
+                    while len(buffer_) > 0:
```
fehiepsi: Hi @jyegerlehner , by changing to this, did you mean that we will pad 0 at the end of the `buffer` if length of `buffer` < length of `sample_size`?

jyegerlehner: Hi @fehiepsi  No, I don't think there's any need for padding. The WaveNet can operate on any length sample. The sample_size just put an upper limit on how big any one sample would be so that memory required could be limited to match available memory.

A problem with the previous implementation is we would never enqueue the sample until it was at least sample_size long. Almost all files are shorter than 100000 samples. So the logic would keep appending wave files to the buffer until it got up to 100000 samples, and then enqueue. And all that after silence trimming. This creates a discontinuity when going from one file's waveform to the next. So we were training it to learn a discontinuity. I think that might be why many of the samples we've been generating have these pops. We've been training the model to create them.

The other reason for the change is that with global conditioning, we must specify a single speaker id when we give the model a waveform to train on. We can't give it a waveform that is an amalgamation of several different files, since each one in general has a different speaker id. Which global id would we use in that case?

jyegerlehner:
> How many steps have you run to get your result?

@fehiepsi I'm sorry to say I don't remember but think it was somewhere in the 20K-30K range. I've since thrown that model away and have been trying various things around scalar_input=True. 

We had some bugs related to scalar input that I've fixed (locally), but still haven't been able to get it to work well yet. It's frustrating because ibab mentioned he got email from DeepMind author confirming that's how they did it, and was something I thought is the "right" way to do it. But as yet not much success.

wavenet/audio_reader.py
fehiepsi: Hi @jyegerlehner , your argument is convincing! The remaining thing I concern is if we have `batch_size=2` for example, then different `sample_size`s will be a problem?

fehiepsi:
I use Python 3.5, and after around 28k steps with `gc_channels=16`, all other settings are default; I get [this](https://soundcloud.com/fehiepsi/speaker362). I cant figure out what is the reason at the moment.

wavenet/audio_reader.py
```diff
@@ -105,14 +169,22 @@ def thread_main(self, sess):
...
+                            sess.run(self.gc_enqueue,
+                                     feed_dict={self.id_placeholder:
+                                                category_id})
                 else:
                     sess.run(self.enqueue,
                              feed_dict={self.sample_placeholder: audio})
+                    if self.gc_enabled:
+                        sess.run(self.gc_enqueue,
+                                 feed_dict={self.id_placeholder:
+                                            categeory_id})
```
fehiepsi: a typo here, this should be `category_id`

wavenet/audio_reader.py
```diff
@@ -1,12 +1,36 @@
...
+            min_id = id
+        elif max_id is None or id > max_id:
+            max_id = id
+
+    return min_id, max_id
+
+
+def randomize_files(files):
+    for file in files:
+        file_index = random.randint(0, (len(files)-1))
```
fehiepsi: pep8: `len(files) - 1`

jyegerlehner:
>  use Python 3.5, and after around 28k steps with gc_channels=16, all other settings are default; I get this. I 

I used gc_channels=32. Maybe train less (because you are overfitting?), or train more (because you haven't trained enough?). Perhaps that's the opposite of a helpful reply; sorry, I don't have a ready answer why your results are different.

jyegerlehner:
> About the scalar input problem, I would like to help. What is the current status?

I've got a model training that has a promising loss profile. I will try generating from it after loss gets a bit lower.

The current ibab master needs a couple fixes: the last two commits to this [branch](https://github.com/jyegerlehner/tensorflow-wavenet/tree/fix-scalar-input), I think.

The latest and greatest I'm training with is [this branch](https://github.com/jyegerlehner/tensorflow-wavenet/tree/fullcorpus_lg). It has a couple other changes in it. If you grab that branch and run `sh traingc.sh` it should duplicate what I'm running at the moment. In case you wanted to do that.

I tried generating from a model a couple days ago that had reasonable loss and just got static. I think I ruled out a gross code error with the following experiment: in `test/test_model.py` I trained a globally-conditioned net with scalar input to generate one of three sine waves where "speaker id" was 0, 1 or 2 to select which sine wave frequency, and it passed the test. I modified the test to save that model to a checkpoint, and then generated from that checkpoint using `generate.py`. I was able to generate each of the desired sine wave frequencies (as specifed by `--gc_id` param). So I think that would seem to rule out an error in the generation code. 

Oh, and fast generation is broken for scalar input. I only use slow generation. That fast generation code is a pain, since it means we need to maintain two duplicate implementations in parallel. 

That's where I'm at, at the moment.

fehiepsi:
Thank you for your explanation! After making some last tests to see if the global condition will work with two different sine wave and to get understanding about why my model fails, I will start to work with scalar input.

jyegerlehner:
@fehiepsi  I found that I have the logs for the samples above. The samples above weren't generated 20K-30K steps in, it was about 146K steps into the training (that was the number of on the checkpoint file right before the wav files were generated).

fehiepsi:
I see. Now it is reasonable. My model seems underfitting. Your model is running through 3 epochs. I will test on just two persons instead. :)

wavenet/audio_reader.py
```diff
@@ -1,12 +1,36 @@
...
+def get_category_cardinality(files):
+    id_reg_expression = re.compile(FILE_PATTERN)
+    min_id = None
+    max_id = None
+    for filename in files:
+        matches = id_reg_expression.findall(filename)[0]
+        id, recording_id = [int(id_) for id_ in matches]
+        if min_id is None or id < min_id:
+            min_id = id
+        elif max_id is None or id > max_id:
```
fehiepsi: bug here: use `if` instead of `elif`

fehiepsi:
I have made the test of global condition on different sine waves which I fake them to be different persons. The overfitting model is doing a real global conditioning. Everything looks good to me.

@jyegerlehner : I will start working on scalar input from now on.

jyegerlehner:
@fehiepsi  On scalar input, I did manage to generate "normal" sounding sample from the branch I mentioned above. That was a first.

jyegerlehner:
> Everything looks good to me.

Thanks for testing it. 

Last call for comments. Anyone object to merging this?

fehiepsi:
My only remaining concern on this is the way we use sample_size to feed data if batch_size > 1. Does it cause any proplem if samples have different sizes when we reshape at various points? In case we use batch_size = 1, if the sample is small then the optimizer will give wrong direction to converge due to 0 padding at the beginning of dilations. If this behavior happens at a large portion of data then the result will be not good.

README.md
```diff
@@ -79,13 +79,30 @@ python train.py --help
...
 
+### Global Conditioning
+Global conditioning refers to modifying the model such that the id of a set of mutually-exclusive categories is specified during training and generation of .wav file.
+In the case of the VCTK, this id is the integer id of the speaker, of which there are over a hundred.
+This allows (indeed requires) that a speaker id be specified at time of generation to select which of the speakers it should mimic. For more details see the paper or source code.
+
+### Training with Global Conditioning
+The instructions above for training refer to training without global conditioning. To train with global conditioning, specify command-line arguments as follows:
+```
+python train.py --data_dir=corpus --gc_channels=16
```
fehiepsi: let `gc_channels=32` for consistence

jyegerlehner:
> My only remaining concern on this is the way we use sample_size to feed data if batch_size > 1. Does it cause any proplem if samples have different sizes when we reshape at various points? 

[TestNetWithGlobalConditioning](https://github.com/jyegerlehner/tensorflow-wavenet/blob/global-conditioning/test/test_model.py#L383) trains the net with a batch size=3, and the test passes. The test covers code in `WaveNetModel`, but not in `train.py`. It's possible there's a defect regarding batch size in the train.py part, but 1) I don't know of it, and 2)  batch size = 1 in all normal use cases.

> In case we use batch_size = 1, if the sample is small then the optimizer will give wrong direction to converge due to 0 padding at the beginning of dilations. If this behavior happens at a large portion of data then the result will be not good.

I'm not sure that's true. Training the net to learn how the time sequence begins (i.e. transitions from silence to speech at the outset of speech) is a valid portion of the training data. We [had a PR to change this](https://github.com/ibab/tensorflow-wavenet/pull/118) and [after some discussion](https://github.com/ibab/tensorflow-wavenet/pull/118) closed it without merging.

> This is another behavior, not about global conditioning. So I suggest to reverse to the original behavior, then merge, and work with this later.

What are you referring to that we should change back to original behavior? I'm not disagreeing, just saying I don't know what change you are suggesting.

fehiepsi:
I agree that we might just use `batch_size=1` so I dont want to dive into the reshape things. About second issue, I have opened an issue where the model is overfiting but still not generate sine wave if we start with silence. By referring to original behavior, I mean that we use `len(buffer) >= sample_size` not `len(buffer) > 0`. Or maybe `len(buffer) >= sample_size / 10` because most audio files in the vctk corpus not have such `sample_size` length.

All of these have nothing to do with `global conditioning`, so no objection to merge. :) I will read your suggested PR tomorrow, it seems interesting.

jyegerlehner:
> By referring to original behavior, I mean that we use len(buffer) >= sample_size not len(buffer) > 0

I discussed the rationale for `len(buffer) > 0` above. I think `len(buffer) >= sample_size` is wrong because it appends > 1 wave file into buffer before enqueueing it for training, which trains the model to learn discontinuities.

>  I have opened an issue where the model is overfiting but still not generate sine wave if we start with silence.

Hmm that's odd. `test/test_model.py` trains to learn 3 different sine waves, conditioned by id 0, 1, 2 for each of three frequencies. There's code in there to save the generated sine waves to .wav files and I've listened to them. It generates fine, and does not use any priming.  Maybe one thing that is different about it is that each training batch starts with a [random duration of silence](https://github.com/jyegerlehner/tensorflow-wavenet/blob/global-conditioning/test/test_model.py#L35) before the beginning of the sine wave. I don't recall if that was necessary to get it to learn or not.

jyegerlehner:
>  I mean that we use len(buffer) >= sample_size not len(buffer) > 0. 

> All of these have nothing to do with global conditioning

The change to `len(buffer) > 0` _is_ related to global conditioning. If we do it the other way, we enqueue a portion of the buffer that is a mix of different wav files from different speakers. Whereas there needs to be a single global id for any audio clip we train on.

fehiepsi:
> I think len(buffer) >= sample_size is wrong because it appends > 1 wave file into buffer before enqueueing it for training, which trains the model to learn discontinuities.

Thank you, I get it. Now I understand what you mean by training the data to learn discontinuity. Though I dont agree with this approach (especially when it is coupled with trimming function), I think that I have to test more before giving further thoughts. :)

fehiepsi:
@jyegerlehner : I have worked on your branch `fix_scalar_input` for a while and see that with your fixes and setting `initial_filter_width=2` (fast generation still not works with filter length > 2), the output of slow and fast generation with `scalar_input` and `global_condition` are the same. I have tested on the corpus of two persons (with setting `initial_sample = [0]` and use `argmax` for the prediction of next sample). Is there any other bug which I am missing?

jyegerlehner:
@fehiepsi 

> fast generation still not works with filter length > 2

That's not particular to this branch.

>  Is there any other bug which I am missing?

I'm not aware of any bugs.

fehiepsi:
So everything is on a right tract now. (I have no intention to consider filter length > 2, just mentioned to emphasize the change of initial filter length for scalar input.) :+1:  for merge your fix-scalar-input branch, change initial filter length (or remove it, because I don't see any reason to keep it distinguish from 'filter length'), and commit. You have been doing a great job! Thank you.

jyegerlehner:
@fehiepsi  As far as I know only lemonzi and ibab have commit privileges. I'm hoping someone who is willing to review PRs would volunteer to referee and ibab might make him a collaborator. In the mean time we wait. 

I have a few more PRs I'd like to submit such as the scalar input fixes, separate test and training losses, but they all are on top of global conditioning and don't want to submit them until that's merged. 

I've got pretty good results with scalar input and global conditioning now.

akademi4eg:
Hey guys, it has been a while since last updates to master. After reading the discussion it looks like this branch is ready-to-use and only awaits merging, am I right?

jyegerlehner:
@akademi4eg  Yes. I've been running it. Along with other changes over in my repo for which I haven't submitted PRs.

lemonzi:
Hi all -- no time for in-depth reviews for the currently open PR's, but since it seems several of you have been using this branch for a while now I'll go ahead and merge it. Thanks a lot @jyegerlehner!

PR #174, opened 2016-11-14
Shuffle (WIP)

nakosung:
Load all batches from audio_reader and shuffle them.

### DIFF
--- a/train.py
+++ b/train.py
@@ -213,8 +213,9 @@ def main():
             coord,
             sample_rate=wavenet_params['sample_rate'],
             sample_size=args.sample_size,
+            batch_size=args.batch_size,
             silence_threshold=args.silence_threshold)
-        audio_batch = reader.dequeue(args.batch_size)
+        audio_batch = reader.dequeue()
 
     # Create network.
     net = WaveNetModel(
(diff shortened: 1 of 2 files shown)


jyegerlehner:
I'm not sure this approach is going to work for us. Am I right you've switched us from a FIFOQueue to tf.train.slice_input_producer in order to exploit its shuffling feature? As coded, it looks like it will take 32 files at a time, and shuffle within those 32 ( or whatever the "batch size" refers to in the tf.train.slice_input_producer [docs](https://www.tensorflow.org/versions/r0.11/api_docs/python/io_ops.html#slice_input_producer)). 

Well the file list tends to give us all 400+ files from one speaker, and then all the files from another speaker as we iterate. So the shuffling that would happen would almost always be within a single speaker, not mixing up the speakers. 

I implemented shuffling with a much smaller change to the code as [a couple lines in a generator](https://github.com/jyegerlehner/tensorflow-wavenet/blob/global-conditioning/wavenet/audio_reader.py#L27) in PR 168. If we merge this PR then a big chunk of PR 168 is going to have to be re-written.

So please sell us on this nakosung, or correct me where I'm wrong please.

nakosung:
As far as I know, the code I wrote is shuffling all tensors(not only small number of slices in a batch). Actually it reads whole dataset first and slices them and feeds via shuffle_batch. But it seems not working because I failed to generate good samples with trained net.

PR #176, opened 2016-11-16
Adding local and global features to AudioReader queue

alexbeloi:
In order to do local/global conditioning, we need the queue manager to enqueue/batch the audio together with the local and global features.

In this PR: AudioReader queue will enqueue (audio, global_features, local_features), if no global/local features are given, they default to zero with the appropriate shape.

### DIFF
--- a/train.py
+++ b/train.py
@@ -214,7 +214,7 @@ def main():
             sample_rate=wavenet_params['sample_rate'],
             sample_size=args.sample_size,
             silence_threshold=args.silence_threshold)
-        audio_batch = reader.dequeue(args.batch_size)
+        audio_batch, gf_batch, lf_batch = reader.dequeue(args.batch_size)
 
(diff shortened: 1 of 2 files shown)


akademi4eg:
With merge of global conditioning PR we have a dilemma.
Option 1. Leave current approach in master branch, i.e. two separate queues for waveforms and global features.
Option 2. Update code to this PR and have one queue for all waveforms, global and local features.

In my opinion, main advantage of option 2 is that it would be harder to accidentally mix up conditions from separate sounds.
On the other hand, it seems more reasonable to merge this kind of change when local conditioning will be ready.

Any thoughts?

lemonzi:
I think having a single queue will help, and it's better to merge small, incremental changes, so I'd say yes if the local conditioning will be ready soon-ish (I lost track of the current state).

Issue #177, opened 2016-11-16
Another issue of model failing to learn a sine wave.


fehiepsi:
In https://github.com/ibab/tensorflow-wavenet/issues/141, the main discussion is about the slightly change in wave form when generating sine wave using quantization 256, which according to me is not the cause of noise. The trimming method is also not a problem because I disabled it in my test.

I tried to make the model learn to create the sin wave, but it turns out that it generates a totally noise audio. Here is a [picture](https://postimg.org/image/i2wc0j9ln/) which I got from the plot `generated_audio[7900:8100]`. I use `seed` of the same sin wave audio file to eliminate the effect of initial seed. I even let the optimizer run through 500 steps to get the `loss=0.0001`, so the model overfits. As you can see from the picture, the first 100 points is the sine wave (a slightly change due to quantization) getting from the seed, but after that, when start to generate, the model fails.

**Edit**: I forget to test the fast_generation method. It can confirm that it is the cause of the problem. Relevant issue is from here: https://github.com/ibab/tensorflow-wavenet/issues/125.

Zeta36:
@fehiepsi , try changing this line in generate.py:

sample = np.random.choice(np.arange(quantization_channels), p=prediction)

by this:

sample = np.argmax(prediction) 

And tell me.

If it doesn't work, change too the line:

waveform = np.random.randint(quantization_channels, size=(1,)).tolist()

by the first float value of the sine wave. For example:

waveform = [.5]

I used this method with a overfitted model and I could replicate the original sample. I did it for example in here #125 and in here #129 (I use a image raw input but it the same)

If all of this fails, you could try this #155 (I know the PR was rejected but I still think it may be the cause of the noise). If you try this you could post here the diferencies between the sine plots you generate with and without the #155 method.

Regards.

fehiepsi:
@Zeta36 , thank you very much for your suggestion. I have discovered the bug and have submitted the fix. For the $\mu$ law encoding, I find it gives us no problem but just a tiny noise.

Zeta36:
@fehiepsi, and did you were able to print in this way a perfect sine plot? could you please link over here the picture of the generated sine wave after the overfitting (and with the bug in the fast generetation fixed)?

fehiepsi:
@Zeta36 : it is not perfect, here is a [picture](https://postimg.org/image/3pcxhwsnn/)

jyegerlehner:
[...]

PR #178, opened 2016-11-16
Fix fast generation bug when using seed

fehiepsi:
This pull request fixes this issue https://github.com/ibab/tensorflow-wavenet/issues/177. I hope that it also fix various issues concerning about generating a good sound.

### DIFF
--- a/generate.py
+++ b/generate.py
@@ -172,7 +172,7 @@ def main():
         outputs.extend(net.push_ops)
 
         print('Priming generation...')
-        for i, x in enumerate(waveform[:-(args.window + 1)]):
+        for i, x in enumerate(waveform[-args.window: -1]):
             if i % 100 == 0:
                 print('Priming sample {}'.format(i))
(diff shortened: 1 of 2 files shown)


generate.py
```diff
@@ -172,7 +172,7 @@ def main():
         outputs.extend(net.push_ops)
 
         print('Priming generation...')
-        for i, x in enumerate(waveform[:-(args.window + 1)]):
+        for i, x in enumerate(waveform[-args.window: -1]):
```
jyegerlehner: So instead of priming with the first args.window samples from the seed waveform, this takes the last args.window samples. Why does that make any difference?

jyegerlehner: Ah, I misunderstood. The old code `waveform[:-(args.window + 1)]` is not taking a portion of size args.window, it's taking everything from the beginning until args.window from the end. Which will be a different size every time, and isn't what we wanted.

PR #184, opened 2016-11-21
Nonlinearity: elu, dropout

nakosung:
Grabbed some nonlinearities from openai/pixel-cnn implementation.

* concat-elu, elu: not to underfit
* dropout: not to overfit


nakosung:
After some trials it seems there is no significant performance boost by applying concat elu, elu and dropout.

PR #203, opened 2017-01-18
Fix a bug preventing training without conditioning

devinplatt:
Hey guys, thanks for creating this great repo! I was very excited that I could train a Wavenet model over the weekend :)

I tried to train a network this weekend on a single large audio file, and I had to edit the code to make it work. It turns out that the findall method of Python's regex returns an empty list when there are no matches, instead of None as was assumed here (if None were returned, the "not None" in the new code would still return True).


devinplatt:
Also, this is my first pull request, so let me know if I did anything incorrect. Thanks!

akademi4eg:
Thanks, @devinplatt !

PR #205, opened 2017-01-19
shuffle input files

weixsong:
This will shuffle all the inputs files and make sure that each input file is used by training process.

Original code could not guarantee that each input file is used by random selection.


weixsong:
@akademi4eg , I think you are right. Different set of training data could help model accuracy.

Issue #209, opened 2017-01-27
The mu-law encoding overflows occasionally


cbquillen:
I tried changing the tf.nn.softmax_cross_entropy_with_logits() for the loss to use tf.nn.sparse_softmax_cross_entropy_with_logits() and then got NaNs.  The problem turned out to be encoded mu-law values > quantization_levels.  It turns out that librosa's load (in audio_reader.py) can produce audio with |audio| > 1.   This isn't all that rare (every 20 chunks or so.)

veqtor:
Could we normalise the audio file using some librosa function?
It's likely that sample-rate conversion causes these types of problems.

PR #210, opened 2017-01-27
Clamp amplitude between -1.0 and 1.0.

jyegerlehner:
This is intended as a fix for issue #209.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,9 +4,9 @@ dist: trusty
 matrix:
     include:
         - python: 2.7
-          env: TENSORFLOW="0.10.0"
+          env: TENSORFLOW="0.12.1"
         - python: 3.5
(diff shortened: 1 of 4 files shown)


jyegerlehner:
I updated to TF r1.0 in case I needed a newer version of TF to see the problem. A half dozen runs under python 2.7 locally don't fail locally, but python 2.7 fail[ed] every time on the CI server.

BTW I also added changes that TF 1.0 is going to require (tf.scalar_summary -> tf.summary.scalar). 

So I give up for the moment until I get a new idea, or someone else has an idea regarding how to reproduce this failure. Maybe someone else can try a PR to fix issue 209 and see if it passes.

[Edit] I see the last failure is in the python 3.5 run, but all the prior ones were in the python 2.7 run. I'm running 2.7 locally.

PR #215, opened 2017-02-07
added default values and one more argument "max_checkpoints" to --help

Fjolnir-Dvorak:
added default values to --help and added a new parameter --max_checkpoints to prevent default deleting checkpoints if there are more than 5.

### DIFF
--- a/train.py
+++ b/train.py
@@ -32,6 +32,7 @@ L2_REGULARIZATION_STRENGTH = 0
 SILENCE_THRESHOLD = 0.3
 EPSILON = 0.001
 MOMENTUM = 0.9
+MAX_TO_KEEP = 5
 
 
 def get_arguments():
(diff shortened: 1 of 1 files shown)


train.py
```diff
@@ -267,7 +270,7 @@ def main():
     sess.run(init)
 
     # Saver for storing checkpoints of the model.
-    saver = tf.train.Saver(var_list=tf.trainable_variables())
+    saver = tf.train.Saver(var_list=tf.trainable_variables(), max_to_keep=args.max_checkpoints) # TODO hier ansetzen
```
akademi4eg: What does this TODO mean?

train.py
```diff
@@ -44,13 +45,13 @@ def _str_to_bool(s):
 
     parser = argparse.ArgumentParser(description='WaveNet example network')
     parser.add_argument('--batch_size', type=int, default=BATCH_SIZE,
-                        help='How many wav files to process at once.')
+                        help='How many wav files to process at once. Default: 1')
```
akademi4eg: This should be taken from BATCH_SIZE variable.

train.py
```diff
@@ -44,13 +45,13 @@ def _str_to_bool(s):
...
     parser.add_argument('--data_dir', type=str, default=DATA_DIRECTORY,
                         help='The directory containing the VCTK corpus.')
     parser.add_argument('--store_metadata', type=bool, default=False,
                         help='Whether to store advanced debugging information '
                         '(execution time, memory consumption) for use with '
-                        'TensorBoard.')
+                        'TensorBoard. Default: False')
```
akademi4eg: Let's also use a variable with default parameter value here (and other similar places).

akademi4eg:
What is a use case where having more than few last checkpoints would be useful?

Fjolnir-Dvorak:
- To be able to perform better bug analysis like locating the appearance of a NaN Loss value on the latest commit from tensorflow-wavenet which happened somewhere (Forgot to grep and pipe the consolte output into a file 'cause I thought the issue were closed. Trying to figure out where the NaN Loss came from).
- To analyse the learning of the net and perform research on ideas to improve the net and evaluate why the net functions as it does and how it evolvet to that.
- Usability, functionality (More features which are pre given and need not to be tested and do not overcomplicate the program are a good thing in my opinion)
- More information to be able to research the human recognition of sounds and the quality and cardinality of the net
- To be able to let the user decide if he only want to have five checkpoints or if he want to do in-depth study.

akademi4eg:
@Fjolnir-Dvorak ok, sounds reasonable.

PR #234, opened 2017-03-06
TF 1.0.0 requires keyword arguments for the loss.

lemonzi:
There's been a long and extended discussion about this minimal change. If the keyword form is backwards compatible (the CI will tell), we should just change it. We might need more changes to make the code fully compatible with TF 1.0.0, though (the official version for this repo is still 0.12).


lemonzi:
The CI seems happy with it and the commit has been approved (thanks @larsr), so merging now! This fixes #223.

PR #239, opened 2017-03-08
Add tensorflow to the requirements

lemonzi:
Will hopefully prevent issues like #238. Future pull requests should use the 1.0.0 API, when a breaking change is introduced we can update this.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,1 +1,2 @@
 librosa>=0.4.3
+tensorflow>=0.12
--- a/README.md
+++ b/README.md
@@ -54,7 +54,7 @@ Code is tested on TensorFlow version 0.12.1 for Python 2.7 and Python 3.5.
 
 In addition, [librosa](https://github.com/librosa/librosa) must be installed for reading and writing audio.
 
-To install the required python packages (except TensorFlow), run
+To install the required python packages, run
 ```bash
 pip install -r requirements.txt
 ```


lemonzi:
Closes #221.

requirements.txt
```diff
@@ -1 +1,2 @@
 librosa>=0.4.3
+tensorflow>=0.12
```
akademi4eg: Maybe tensorflow-gpu would be more reasonable?
Most of CPU tensorflow users might not want to train this model anyway.

lemonzi: Does pip complain if you try to install tensorflow when tensorflow-gpu is already installed? (tensorflow's pip doesn't work for me and I don't have a GPU, so I can't test it). If that's the case, leaving the CPU version would mean that people just downloading the repo and testing it won't have issues and won't complain about cuda errors, and I assume people who have a decent enough GPU already have tensorflow or know how to get it. If pip allows installing both of them and then everything breaks, it would probably be better to use the gpu version. But the CI doesn't have GPU, so we'd have to adapt the script.

lemonzi: I did some searching and dozens of repos include tensorflow in their requirements but I only found one with tensorflow-gpu, which suggests having tensorflow-gpu already installed just ignores the regular tensorflow. It would be good to double-check though.

PR #241, opened 2017-03-08
TensorFlow 1.0

lemonzi:
Upgrade to TensorFlow 1.0. No breaking changes introduced so far.

PR #248, opened 2017-03-25
Fix SummaryWritter error and similar deprecated warnings

mchinen:
TF1.0 advises to use FileWriter and similarly pathed functions to replace deprecated ones.
  All changes here are related to fixing deprecated errors and warnings.


lemonzi:
Official migration to TensorFlow 1.0 is being discussed in #241. Since these changes are compatible with 0.12 though (Travis hasn't complained), I'll go ahead and merge them.

PR #327, opened 2018-02-08
Remove declarations for unused variables

isnbh0:
In `_create_network` in model.py, There was an unnecessary conditional declaration of
`initial_channels`, a variable unused elsewhere within the method. Likely
left over from previous modifications. I removed the declaration.


isnbh0:
resubmitted at #329.

PR #352, opened 2018-07-24
CPU optimization: adding support of threading and profiling

ashraf-bhuiyan:
In this PR, intra-op and inter-op threading are supported in the model. Moreover, profiling and new timer is also added. Readme file is also updated to show how to use intra-op and inter-op threading to get great performance on CPUs.


ashraf-bhuiyan:
Can you please take a look at this PR?

PR #383, opened 2019-08-14
Root mean square error (rmse) has been renamed to rms. from librosa b…

kshina76:
Root mean square error (rmse) has been renamed to rms from librosa ver 0.7.


kshina76:
in audio_reader at 68 lines.

Issue #391, opened 2020-02-07
tensorboard result: the generated audio of generate.py is 0 seconds


Ahapy:
Environment: win10 + Python3.7, cuda10.2:
1. Copy part of the audio of VCTK / wav48 / to form a wav folder for training
2.New logdir
3. Modify "SILENCE_THRESHOLD" in the original program to 0.1
tensorboard result: the generated speech of generate.py is 0 seconds

Issue #338, opened 2018-04-04
How to use tensorboard


zsxgb:
My tensorboard is blank.

I use conda, not tensorflow installed using pip. I thought at first that it was my tensorflow and tensorboard version issues. I dealt with it for a long time until I successfully started tensorboard on another tensorflow program. I found that it is not a version issue.

What should I do with this issue, this program must be running in version 1.0.0? It seems that now the conda minimum version of tensor is 1.0.1, and I should

zsxgb:
![2018-04-04 11-12-14](https://user-images.githubusercontent.com/14177065/38286679-26b69392-37f9-11e8-94d9-111500e9c091.png)
![2018-04-04 11-13-27](https://user-images.githubusercontent.com/14177065/38286692-39e63f58-37f9-11e8-92b9-757cc7549e05.png)

Benimation:
I can see the graphs while/after training, but I can't playback the sounds after generating. It always shows 3 audio players, but none of them work.

Issue #262, opened 2017-04-25
Empty Generated audio


Deejep:
Hi,

I have trained wavenet on 109 files of hi quality french audio (no noise, each file is approx 5 seconds).
Unfortunately after 70K steps of training, I got files (called individual) that don't play out from tensorboard.
PS: I have set the silence threshold to 0 because of the hi quality.

Tunas1337:
Has anyone ever found a solution to this?

Issue #232, opened 2017-03-01
Cannot see anything from Tensorboard


BuchyLX:
Hi,

I ran the script while set --histogram=true. Then I got the "events.out.tfevents..." file in the logdir. The size of it is about 5G.
But when I started the tensorboard and I got nothing from the websit.
It looks like the screen-cut I attached.

Anyone can help me with this problem?
![tensorboard](https://cloud.githubusercontent.com/assets/17761044/23455062/5837eb4e-fea9-11e6-9002-447f4add441b.png)

BuchyLX:
I use Chrome rather than Safari then it works. 
Er...But how to close the issue?

Deejep:
Just click on the AUDIO tab.

ChrisChan2013:
hi, I also cannot get anything from Tensorboard after running the VCTK-corpus. There is only zero second record in Audio. What should I do next?

PR #397, opened 2020-08-12
Fix generate.py to write summary_out properly

ghleokim:
Hello,

I am trying to generate audio with this model.

Giving `wav_out_path` properly, the output seems to be generated as intended. However, when I run a tensorboard server with logdir/ files, the summary file does not show up properly.

I looked up the issue tab on this repo. As mentioned in several other issues like #391, #338, #262, and #232, I couldn't see generated output. My case was the same.

By changing the shape inside `summary_out(...)` from `[-1,1]` to `[1,-1]` solved this problem.

### DIFF
--- a/generate.py
+++ b/generate.py
@@ -264,7 +264,7 @@ def main():
     tf.summary.audio('generated', decode, wavenet_params['sample_rate'])
     summaries = tf.summary.merge_all()
     summary_out = sess.run(summaries,
-                           feed_dict={samples: np.reshape(waveform, [-1, 1])})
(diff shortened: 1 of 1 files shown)


ghleokim:
1. Tensorboard output with original code

Screen Shot 2020-08-12 at 4 40 17 PM

2. Tensorboard output with modified code
Screen Shot 2020-08-12 at 4 40 27 PM

3. Also double-checked if the shown audio is identical to the generated output file.
Screen Shot 2020-08-12 at 4 41 10 PM

